PR #13, opened 2019-11-22
remove need to clone

kalantar:
I've updated the documentation to remove the need to clone the repositories. However, it relies on the addition of a yaml file for install of iter8-analytics for knative since this is different from the install for istio.

### DIFF
--- a/doc_files/iter8_install.md
+++ b/doc_files/iter8_install.md
@@ -102,6 +102,4 @@ If you want to uninstall all _iter8_ components from your Kubernetes cluster, fi
 kubectl delete -f https://raw.githubusercontent.com/iter8-tools/iter8-controller/master/install/iter8-controller.yaml
 ```
 
-Note that this command will delete the `Experiment` CRD and wipe out the `iter8` namespace.
-
-Note that this command does not remove the Grafana dashboard if created.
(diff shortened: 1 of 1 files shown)


kalantar:
Addresses https://github.com/iter8-tools/docs/issues/11

kalantar:
prereq for this is https://github.com/iter8-tools/iter8-analytics/pull/23

fabolive:
@kalantar: I pushed a few enhancements to your branch here. Please, take a look.
Also, I have a comment. Right now, we ask the users to import the Grafana dashboard, which can only be done if one clones the repo(s). Could you, please, modify the installation instructions to tell the users to run your script https://github.com/iter8-tools/iter8-controller/blob/master/hack/grafana_install_dashboard.sh to import the dashboard? We can tell the users to manually import if they cannot run the script properly.

kalantar:
Good idea. Currently it assumes the config file is local. I think we will need to modify the script to get it from the repository instead.

kalantar:
Updated `grafana_install_dashboard.sh` (to allow use of URL when specifying the config file) is defined by PR https://github.com/iter8-tools/iter8-controller/pull/49. It should be mreged before this one.

kalantar:
@fabolive Please take another look

PR #33, opened 2020-03-25
Update doc

Alan-Cha:
Change gateway URLs so they are in the same format


kalantar (approved):
Good catch

kalantar:
@fabolive I don't have permission to merge code for this project

PR #34, opened 2020-03-25
Fix typo in docs

Alan-Cha:
@kalantar oops, so sorry about that. I was a little careless. Good thing I checked again.


kalantar (approved):
@Alan-Cha I should have caught this

PR #85, opened 2020-05-22
openshift specific instructions

kalantar:
Red Hat specific install documentation.

@fabolive I've already pushed to v1.0 branch but we should consider adding to the v0.1 release. It will need to be linked from somewhere. I leave it to your discretion ...


kalantar:
Should have similar documentation for v1.0.0

kalantar:
changes also pushed to master for v1.0.0

PR #133, opened 2020-07-10
fix install on redhat instructions

kalantar:


kalantar:
@Alan-Cha I've updated the markdown files but the new html file should really be the one updated...

PR #162, opened 2020-07-17
A number of improvements

Alan-Cha:


Alan-Cha:
@sriumcp Agreed. That's why I want to look into deploying a React app instead but I'm having trouble with the styling.

PR #216, opened 2020-08-13
Miscellaneous changes

Alan-Cha:


Alan-Cha:
@kalantar I also changed Netlify's build command so upon accepting any new change, the broken links in the tutorial should be fixed again.

PR #241, opened 2020-08-14
Replicate some of the changes in v1.0.0 preview

Alan-Cha:


Alan-Cha:
@kalantar I reverted back to the old page. Something to note is that I did change the title of the about page to "About" because the old title was just too long and jarring. I made the old title a subtitle.

PR #250, opened 2020-08-17
instructions to get unstuck when namespace cannot be deleted

huang195:


content/installation/kubernetes.md
```diff
@@ -65,3 +65,16 @@ If you want to uninstall all of iter8 components from your Kubernetes cluster, f
 ```bash
 kubectl delete -f https://raw.githubusercontent.com/iter8-tools/iter8-controller/v1.0.0-preview/install/iter8-controller.yaml
 ```
+
+### Uninstall is stuck?
+
+Iter8 uses K8s finalizer to ensure things are cleaned up properly. However, if
+one removes *iter8-controller*, e.g., by running the above uninstall command
+before the workload namespace is cleaned up, removing the workload namespace
+afterward could get stuck. To unstuck, one would need to manually edit the
```
kalantar, line 74: Perhaps modify as: "one would need to manually edit (with `kubectl edit`) the ..."

huang195: Revised

PR #269, opened 2020-08-19
Algorithms (with traffic control description)

sriumcp:


kalantar:
I’ve approved the PR. It is on the master branch and will feed http://preliminary.iter8.tools
However, we have created the branch `release-1.0.0` (for release v1.0.0-rc1) which feeds http://iter8.tools. This means that in addition to applying your change to the master branch, you should create a second PR to add the change to the release-1.0.0 branch.  To do so, you can use git cherry-pick to create a branch with the new commits.

PR #319, opened 2020-09-14
updated experiment doc for networking changes

kalantar:


content/reference/experiment.md
```diff
@@ -83,9 +74,34 @@ service:
...
+*gateway* | string | The name of gateway to which this host is attached. | yes
+
+An example of the `networking` subsection of an experiment object is as follows:
+
+```yaml
+  networking:
+    hosts:
+      - name: bookinfo.example.com
+        gateway: bookinfo-gateway
+  id: bookinfo
```
kimikowang, line 104: `id` is out of its parent section.

kimikowang:
Is it possible to remove the type column from each table? I think that might be unnecessary.

content/reference/experiment.md
kalantar: Good catch. Fixed.

kalantar:
> Is it possible to remove the type column from each table? I think that might be unnecessary.

I suppose so. I find it helpful for some of the later items especially when there is an enum.

sriumcp:
On types, here's how Istio documents CRDs: https://istio.io/latest/docs/reference/config/networking/virtual-service/. Experiment crd description is directly inspired by Istio (table format here = Istio's table format).

kimikowang:
> On types, here's how Istio documents CRDs: https://istio.io/latest/docs/reference/config/networking/virtual-service/. Experiment crd description is directly inspired by Istio (table format here = Istio's table format).

Those types are hyperlinks. But we do not have links for the types in the doc.

kimikowang (approved):

sriumcp:
Some of those types are hyperlinks in Istio docs. Not all (for e.g., there are strings and string arrays in types without types).

Yes, a PR with hyperlinks in types (where they exist) would be a nice addition.

PR #326, opened 2020-09-16
Add link checker

Alan-Cha:


sriumcp (approved):
LGTM. Does this find broken links in doc automagically?

Alan-Cha:
@sriumcp Yes, it is able to find broken links in the docs "automagically" 🤣 It'll run whenever someone pushes or files a PR. 

It is currently ignoring `http://iter8-analytics.iter8:8080` (which is used in a tutorial) and URLs with the pattern `.*releases/download/master/.*`, because `master` is not a proper release tag and these URLs are autogenerated using the short code the Michael and I create a few days ago. Regarding the URLs with the pattern, do we want to make sure that all links in preliminary.iter8.tools works? Because currently, they do not. I think this is an extended conversation that is probably better suited in a voice call.

PR #331, opened 2020-09-17
Wording change

Alan-Cha:
The name of the step that is being referenced is "Install Kiali Latest", not "Install Kiali"


sriumcp (approved):
LGTM.

PR #334, opened 2020-09-18
Fix search

Alan-Cha:
The problem seems to stem from a using a different version of jquery. I reverted back to the original version.


Alan-Cha:
And the link checker action worked too!

Alan-Cha:
Wait it stopped working again... hold on

Alan-Cha:
@sriumcp It works locally but not in the preview. Could you clone my fork and try it locally as well? Sometimes it seems that the preview doesn't always work as intended...

sriumcp (approved):
LGTM.

PR #341, opened 2020-09-21
Fix search

Alan-Cha:
I accidentally pushed the `fix search` commit to the `release-1.0.0` branch. This is because my branch was pointing to the wrong remote. I have fixed it but should we also consider making these branches protected?


sriumcp (approved):
LGTM.

PR #397, opened 2020-10-08
add -L option to curl

kalantar:


kalantar:
@Alan-Cha the failure seems to be in the link to the RH doc. I've checked using the red doc search itself and these links are gone. I imagine this means that they are updating them. I think we should merge and recheck the links tomorrow.

PR #405, opened 2020-10-15
Add roadmap

Alan-Cha:


sriumcp (changes requested):
Perhaps this line can be removed? `Learn about what we are working on and what we are planning to work on.` 

Or replaced with something along the following lines? "Here are some of the iter8 enhancements coming soon...".

PR #399, opened 2020-10-08
document resuse of vs

kalantar:
@sriumcp @sushmarchandran since you had the recent experience of working with a user who had this problem ... does this help?

PR #410, opened 2020-10-22
document resuse of vs (#399)

kalantar:
* document resuse of vs

* fix spelling of VirtualService


kalantar:
@Alan-Cha I discovered I cherry-picked and prepared this but did not push and create the pull request. Here it is for v1-0-0

PR #457, opened 2021-02-23
E2e tests added for knative

sushmarchandran:


sriumcp (changes requested):
The changes look good... however, I have a `themes/hugo-themes-learn` showing up in the list of changes. It says, `Submodule hugo-theme-learn` added. Could you please get rid of this from the commit?

PR #546, opened 2021-03-17
deleting gh banner

sriumcp:


sriumcp (commented):
removing ghbanner so that it can be added again.

PR #568, opened 2021-03-29
Rolebinding

sriumcp:
Previous e2e tests were not checking for experiment conditions properly... 

This PR fixes all the e2e tests to ensure tests are run properly (both the 4 automated ones, and the 2 manual ones).


kalantar:
It seems like the checks are repeated many times. Consider creating a "library.sh" of checks. It would be "loaded" using `source library.sh` and then each individual list of checks can then be simplified. We might even want to have a general library (for things like checking request loss (fortio) and completion (experiment) and a specific library for knative (to check things like traffic distribution) that can be replaced for different domains.

sriumcp:
@kalantar Couldn't agree more with your comments above -- my preference is to move away from shell scripts to, say Python. It is of course possible to create library of shell script functions -- just extremely painful and error prone, and not worth it.

sushmarchandran (approved):
lgtm

kalantar:
@sriumcp there is no rush on my suggestion and it can wait until the next iteration should we find ourselves still using shell scripts.

kalantar (approved):

kalantar (approved):

PR #579, opened 2021-04-01
Remove metric details

sriumcp:
A lot more work remains to be done in order to describe Iter8's custom metrics story... removing relevant documents for now since they are incomplete (and in many places, in correct)


kalantar (approved):
lgtm

PR #590, opened 2021-04-07
set up  versioning

sriumcp:
This PR addressed step 1 documented in this issue: #591 

This PR enables versioning for `Material for MkDocs` as documented here: https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/

Since we're using Material for MkDocs, the idea is to simply use the versioning method recommended and documented by this tool.

### DIFF
--- a/mkdocs/mkdocs.yml
+++ b/mkdocs/mkdocs.yml
@@ -47,6 +47,9 @@ plugins:
 # Customization
 extra:
   title: "Cloud native release automation and experimentation"
+  extra:
+    version:
+      provider: mike  
   social:
     - icon: fontawesome/brands/github
       link: https://github.com/iter8-tools/iter8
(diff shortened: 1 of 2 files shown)


kalantar:
The documentation suggests setting a default version and a warning if not viewing the default version. Are these steps planned as future steps? Both seem like a good idea.

sriumcp:
Setting a default versioning is possible and will be part of steps 2 and 3.

I believe "Warning if not viewing the default version" is currently available to sponsors only. We will absolutely make use of this feature once it becomes available widely to the general public in Material for MkDocs.

There are a couple of other related features, like being able to indicate Git tags as well next to the repo -- also currently for sponsors only.

kalantar (approved):

PR #602, opened 2021-04-10
Upgrade to Material for MkDocs - 7.0

sriumcp:
This is needed in order to enable navigation across different versions of Iter8 in https://iter8.tools.


sriumcp (commented):
looking  alright.

PR #628, opened 2021-04-22
Istio quickstart

### LINKED ISSUE #619: [Feature Request] Quickstart for Istio
**Why is the feature useful for Iter8 users?**
Demonstrate how Iter8 might be used for experiments on applications that are deployed using the Istio service mesh.

Quickstart similar in nature to the one provided for knative demonstrating a simple canary experiment.

**Describe alternatives you've considered**
None

No

[ ] TBD

[ ] Add quickstart page

kalantar:
Resolves #619 

Creates quickstart for Istio with an A/B test 
Includes e2etest.sh that executes the same commands.


kalantar:
I found it very difficult to set short timeouts. I found that creating pods could take a few seconds to 10 minutes.

kalantar:
Two items might be considered as prerequisites for this. Opinions welcome

- rewards section of iter8ctl output (issue https://github.com/iter8-tools/iter8ctl/issues/28)
- document A/B as an example testing pattern

sriumcp:
@kalantar I think iter8ctl output need not be considered a pre-req.

I do believe A/B and A/B/n testing pattern documentation is a pre-req.

sriumcp (commented):
1. mkdocs/docs/images/istio-canary-progressive-kubectl-iter8.png -- Could we please keep this file out for now. Since  this doesn't seem relevant to quick start.
2. I agree with your suggestion that checking the source pptx here is a good idea -- could you please do so (I mean the smaller pptx which only has the images relevant to Iter8 docs; not the Kubecon  preso, etc.)

samples/istio/quickstart/experiment.yaml
```diff
@@ -0,0 +1,66 @@
...
+    deploymentPattern: Progressive
+    actions:
+      # when the experiment completes, promote the winning version using kubectl apply
+      finish:
+      - task: common/exec
+        with:
+          cmd: /bin/bash
+          args: 
+          - "-c"
+          - kubectl -n {{ .namespace }} apply -f {{ .promote }}
```
sriumcp: Replacing {{ .namespace }} with the string bookinfo-iter8 will simplify line 21, eliminate the need for lines 42, 43, 57, and 58. I feel this simplicity is desirable.

samples/istio/quickstart/experiment.yaml
```diff
@@ -0,0 +1,66 @@
...
+      upperLimit: "0.01"
+  duration: # product of fields determines length of the experiment
+    intervalSeconds: 10
+    iterationsPerLoop: 10
+  versionInfo:
+    # information about the app versions used in this experiment
+    baseline:
+      name: A
+      variables:
+      - name: revision # used in Prometheus queries
```
sriumcp: `revision` is more of a Knative terminology. Perhaps `subset` or `deployment` is a better term for Istio?

samples/istio/quickstart/experiment.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+  criteria:
+    rewards: # metrics to be used to determine the "value" or "benefit" of a version
+    - metric: books-purchased
+      preferredDirection: High
+    objectives: # metrics to be used to determine validity of a version
+    - metric: iter8-istio/mean-latency
+      upperLimit: 100
+    - metric: iter8-istio/error-rate
+      upperLimit: "0.01"
```
sriumcp, line 29: Explicitly adding request-count is a good idea in all experiments.

samples/istio/quickstart/fortio.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  template:
+    spec:
+      volumes:
+      - name: shared
+        emptyDir: {}    
+      containers:
+      - name: fortio
+        image: fortio/fortio
+        command:
```
sriumcp: Using the square bracket notation as in line 34 will shorten this deployment significantly.

samples/istio/quickstart/fortio.yaml
```diff
@@ -0,0 +1,38 @@
...
+          - fortio
+          - load
+          - -t
+          - 6000s
```
sriumcp: 6000s looks good.

samples/istio/quickstart/bookinfo-gateway.yaml
```diff
@@ -0,0 +1,67 @@
...
+      weight: 100
+    - destination:
+        host: productpage
+        port:
+          number: 9080
+        subset: productpage-v3
+      weight: 0
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: DestinationRule
```
sriumcp, line 54: **Ignore:** I see productpage-v1, and v3, but no v2. Is there a reason? Renaming v3 to v2 will make things more straightforward.

sriumcp (changes requested):

samples/istio/quickstart/B.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  labels:
```
sriumcp: The labels section (which looks like a remnant from `iter8-istio`) seems to add lines to this VS without any purpose.

samples/istio/quickstart/B.yaml
```diff
@@ -0,0 +1,33 @@
...
+    iter8-tools/role: stable
+  name: bookinfo
+spec:
+  gateways:
+  - mesh
+  - bookinfo-gateway
+  hosts:
+  - productpage
+  - "bookinfo.example.com"
+  http:
```
sriumcp, line 12: **Ignore:** The match section can be simplified based on the actual URLs used. The traffic generation is not manual, but  automated... so we can focus on the URI which is actually used by Fortio (instead of listing all).

sriumcp (changes requested):

Makefile
```diff
@@ -0,0 +1,5 @@
+test:
+	for test in $(shell find . -name e2etest.sh -print); do \
+		echo "\033[0;32mRUNNING:\033[0m $$test"; \
```
sriumcp, line 3: I like the fancy echoes 👍

samples/istio/quickstart/B.yaml
fabolive: @sriumcp: Not listing the URLs will break the application, which will show if someone tries to access it from the browser. There might be other consequences to the downstream calls made by productpage.

samples/istio/quickstart/bookinfo-gateway.yaml
fabolive: @kalantar, @sriumcp: Below is a reminder of the behavior of the 3 versions, according to the images we have built:

productpage-v1:
* no errors
* normal latency behavior (no injected delay)
* comparatively low reward

productpage-v2: 
* no errors
* higher latency (some noticeable delay injected)
* reward much much higher than that of v1 and v3

productpage-v3: 
* no errors
* normal latency behavior
* reward higher than v1 but lower than v2

fabolive (commented):
@kalantar: I agree with most comments made by @sriumcp. I only diverge from the suggestion that breaks the application (omitting most URLs in the VS).

sriumcp (commented):
Updated review  comments.

samples/istio/quickstart/experiment.yaml
kalantar: namespace is needed for the metrics we have defined; it will be necessary in the versionInfo anyway.

samples/istio/quickstart/experiment.yaml
kalantar: Making this change has an impact on the metrics definitions. This works for both deployments and services. I would prefer a word that is neutral. I'll use "version".

samples/istio/quickstart/bookinfo-gateway.yaml
kalantar: I left the version names we used from our A/B/n  tutorial in iter8 1.0.0. But I agree this is probably a good change to make.

samples/istio/quickstart/experiment.yaml
sriumcp: Yes, using `namespace` variable within metrics is a good idea. 

However, {{ .namespace }} placeholder in line  21 above doesn't serve any purpose and only makes the kubectl apply command needlessly cryptic, since  {{ .namespace }} is not a dynamically interpolated variable.

kalantar:
The updates now depends on https://github.com/iter8-tools/iter8-install/pull/71

kalantar:
I left the virtualservice as it was. It should be changed as little as possible from the way it defined by Istio.

sriumcp (changes requested):
@kalantar I am making several changes for the Knative tutorials (and therefore, e2e tests) -- mainly as a follow up to the RBAC minimization changes.

Could you please restrict this PR to only Istio related changes and perhaps save the Knative changes for a separate PR?

mkdocs/docs/getting-started/quick-start/with-istio.md
```diff
@@ -1,9 +1,390 @@
...
+        deploymentPattern: Progressive
+        actions:
+          # when the experiment completes, promote the winning version using kubectl apply
+          finish:
+          - task: common/exec
+            with:
+              cmd: /bin/bash
+              args:
+              - "-c"
+              - kubectl -n {{ .namespace }} apply -f {{ .promote }}
```
sriumcp: This line needs to be updated as well.

sriumcp (changes requested):

mkdocs/docs/getting-started/quick-start/with-istio.md
kalantar: done

sriumcp (approved):
lgtm

PR #671, opened 2021-05-06
Deployment patterns

sriumcp:


sriumcp:
To be refactored and handled separately.

PR #736, opened 2021-06-06
Add sample quickstart for Seldon

ukclivecox:
Adds an initial Seldon quickstart using 2 Iris models.

Notes

 * The URLs for promotion in the experiment are set to what they would be when this PR is merged. So if tested would need changing to my branch to work.
 * The platformsetup.sh needs Helm

If merged I can create docs in a lter PR to update quickstart.

### DIFF
--- a/install/core/rbac/stacks/iter8-seldon/kustomization.yaml
+++ b/install/core/rbac/stacks/iter8-seldon/kustomization.yaml
@@ -1,0 +1,3 @@
+resources:
+- rolebindings.yaml
+- roles.yaml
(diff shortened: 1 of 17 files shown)


sriumcp:
@cliveseldon The PR is looking good as is. I will merge it since it includes an e2e test which I can exercise after merging, and also doesn't seem to have upset any of the existing tests.

Very much looking forward to the follow up PR with instructions on how to run this.

sriumcp (approved):
lgtm

Issue #782, opened 2021-06-12
Metrics not found when deleted and recreated with experiment


ukclivecox:
**Describe the bug**
Run an experiment with metrics once. Delete and rerun immediately. Sometimes it will fail with metric not found.

**To Reproduce**

Create Kind cluster
```
kind create cluster
```

Setup platform

```
cd ${ITER8}/samples/seldon/quickstart
./platformsetup.sh
```

Run code below which start experiment/metrics 3 times in a row.

```
#!/bin/bash

set -e

export EXPERIMENT=quickstart-exp

echo "Create app/ML model versions"
kubectl apply -f $ITER8/samples/seldon/quickstart/baseline.yaml
kubectl apply -f $ITER8/samples/seldon/quickstart/candidate.yaml
kubectl apply -f $ITER8/samples/seldon/quickstart/routing-rule.yaml
kubectl wait --for condition=ready --timeout=600s pods --all -n ns-baseline
kubectl wait --for condition=ready --timeout=600s pods --all -n ns-candidate

echo "Generate requests"
URL_VALUE="http://$(kubectl -n istio-system get svc istio-ingressgateway -o jsonpath='{.spec.clusterIP}'):80"
sed "s+URL_VALUE+${URL_VALUE}+g" $ITER8/samples/seldon/quickstart/fortio.yaml | sed "s/6000s/600s/g" | kubectl apply -f -
pod_name=$(kubectl get pods --selector=job-name=fortio-requests -o jsonpath='{.items[*].metadata.name}')
kubectl wait --for=condition=Ready pods/"$pod_name" --timeout=240s
pod_name=$(kubectl get pods --selector=job-name=fortio-irisv1-rewards -o jsonpath='{.items[*].metadata.name}')
kubectl wait --for=condition=Ready pods/"$pod_name" --timeout=240s
pod_name=$(kubectl get pods --selector=job-name=fortio-irisv2-rewards -o jsonpath='{.items[*].metadata.name}')
kubectl wait --for=condition=Ready pods/"$pod_name" --timeout=240s

echo "Define metrics"
kubectl apply -f $ITER8/samples/seldon/quickstart/metrics.yaml

echo "Launch experiment"
kubectl apply -f $ITER8/samples/seldon/quickstart/experiment.yaml


# Wait a few seconds
sleep 10

kubectl delete -f $ITER8/samples/seldon/quickstart/experiment.yaml
[...]

PR #784, opened 2021-06-14
timing related tests

sriumcp:

Reproduces the bug in #782 and shows that the quick hack documented in the discussion works!


samples/e2e-tests/race-condition/platformsetup.sh
```diff
@@ -0,0 +1,38 @@
...
+KUSTOMIZE_VERSION=$(kustomize  version | cut -d. -f1 | tail -c 2)
+if [[ ${KUSTOMIZE_VERSION} -ge "3" ]]; then
+    echo "Kustomize v3+ available"
+else
+    echo "Kustomize v3+ is unavailable"
+    exit 1
+fi
+
+# Step 1: Install Iter8
+echo "Installing Iter8 with Knative support"
```
kalantar: This isn't specific to knative. Change to just Iter8.

kalantar (changes requested):
minor change

samples/e2e-tests/race-condition/platformsetup.sh
sriumcp: Done.

kalantar (approved):

PR #799, opened 2021-06-22
Linkerd support

Alan-Cha:


huang195:
@Alan-Cha summarizing our discussion on Slack here. The issue that's blocking me from testing this locally is the `linkerd viz` components not getting properly installed. I've tried it twice, and it failed both times. I used minikube (a new cluster each time) to test. Earlier, I got the impression that this would "sporadically" fail, thus was recommending you to document the potential failure. However, if it's failing this frequently, we should address it more thoroughly as users won't benefit much from this tutorial if they cannot get linkerd installed.

Alan-Cha:
See #832

PR #807, opened 2021-06-24
do not change deployment label

huang195:
Deployment label is an immutable field. Changing it across versions will result in ArgoCD failing to sync due to it calling `kubectl apply` or `kubectl replace`, for which immutable fields cannot be updated.


sriumcp (approved):
lgtm

Issue #733, opened 2021-06-04
[Feature Request] Bash task to replace/augment common/exec


kalantar:
**Is your feature request related to a problem? Please describe.**
A common use for the common/exec task is to set `cmd: bin/bash` and to have the first argument be `-c`. Instead, provide a task that takes just a script as an argument and runs it as a script. This simplifies expression for the user.

In addition, we can eliminate the need for disableInterpolation; we can determine if versionRecommendedForPromotion has been set or not as part of the processing.

**Why is the feature useful for Iter8 users?**
Simpler expression. Very common pattern in the task we have created thus far.

**Describe the solution you'd like**
Introduce `common/bash` as a new task. It would allow expression as:

      - task: common/bash
        with:
          script: | 
            kubectl -n bookinfo-iter8 apply -f {{ .promote }}

**Describe alternatives you've considered**
There are other options for naming such as common/run. I think being explicit that it is a bash script is probably helpful to the user.

**Does this issue require a design doc/discussion? If there is a link to the design document/discussions, please provide it below.**
This issue can be for design. Propose to add the task as an experimental task. That is, add it. But not necessarily document it. We can use it, get feedback and refine it (or throw it away).

**How will this feature be tested?**
Unit tests for creation, validate any interpolation changes.

**How will this feature be documented?**
Initially, not documented. If promoted from experimental, would document as part of references tasks.

sriumcp:
@kalantar {{ .promote }} still comes from version recommended for promotion. Is this correct?

Also, what happens if this task is invoked and there is no version recommended for promotion in the experiment?

sriumcp:
I feel this is a significant improvement (in terms of user experience, ease of use and readability) over the existing `common/exec` task.

kalantar:
> @kalantar {{ .promote }} still comes from version recommended for promotion. Is this correct?
> 
> Also, what happens if this task is invoked and there is no version recommended for promotion in the experiment?

Good question. The second merits some discussion.

At to the first, yes. {{ .promote }} will come from recommended version. The code now creates the list of tags as follows:

	tags := base.NewTags().
		With("this", obj).
		WithRecommendedVersionForPromotion(&exp.Experiment)

In this case, `obj` is the experiment converted to a map[string]interface{}. Any variables from the version recommended for promotion are also added.   If there is no version recommended for promotion, these are not added.

[...]

PR #829, opened 2021-06-30
documentation for common/bash task

kalantar:

Documentation for common/bash task; issue https://github.com/iter8-tools/iter8/issues/733


mkdocs/docs/reference/tasks/interpolation.md
```diff
@@ -0,0 +1,26 @@
...
+  with:
+    script: |
+        kubectl apply -f {{ .promoteManifest }}
+```
+
+`{{ .promotionManifest}}` is a placeholder.
+
+Placeholders are specified using the Go language specification for data-driven [templates](https://golang.org/pkg/html/template/). In particular, placeholders are specified between double curly braces.
+
+Iter8 supporsts placeholders for:
```
sriumcp: Fix spelling.

mkdocs/docs/reference/tasks/common.md
```diff
@@ -61,20 +115,8 @@ In the [example above](#example), a YAML file corresponding to the baseline or c
...
-kubectl apply -f https://raw.githubusercontent.com/iter8-tools/iter8/master/samples/knative/quickstart/{{ .promote }}.yaml
-```
-In this case, the placeholder is `{{ .promote }}`. 
-
-Placeholder substitution in task inputs works as follows. 
-
-Iter8 will find the version recommended for promotion which is determined by Iter8 during the  course of the experiment, and stored in the `status.versionRecommendedForPromotion` field of the experiment resource. The version recommended for promotion is the winner, if a winner has been found in the experiment. Otherwise, it is the baseline version supplied in the `spec.versionInfo` field of the experiment.
-
-If the placeholder is `{{ .name }}`, Iter8 will substitute it with the name of the version recommended for promotion. If it is any other variable, Iter8 will substitute it with the value of the corresponding variable for the version recommended for promotion. Variable values are specified in the variables field of the version detail when the experiment is created.
+### Limitations
```
sriumcp: The deprecation is fine. However, I think the "limitations" title is incorrect. `disableInterpolation` flag is a feature, not a limitation for instance :) IMO, the only limitation of `exec` over `bash` is its verbosity. `bash` will cut out a line or two in many situations, which is certainly a very welcome type of simplicity.

Also, when there is no version recommended for promotion, but interpolation is required, the two tasks behave differently (in the case of `exec`, it will fail, and in the case of `bash` it will quietly output something which is probably not the desirable output anyway).

mkdocs/mkdocs.yml
```diff
@@ -173,6 +173,7 @@ nav:
       - Common Tasks: reference/tasks/common.md
       - Metrics Tasks: reference/tasks/metrics.md
       - Notification Tasks: reference/tasks/notification.md
+      - Dynamic Variable Substitution: reference/tasks/interpolation.md
```
sriumcp: `mkdocs` does not feel like the the right place for this link. This link needs to be part of `tasks.md` file instead.

For starters, `interpolation` is not a separate task and is the exception in that list.

I actually anticipate other tasks (especially GitOps) tasks being written, where the interpolation behavior can differ from everything we have done so far.

sriumcp (changes requested):
see comments.

mkdocs/mkdocs.yml
kalantar: Moved as suggested. I hope that we will figure out how to extend/refactor the notion instead of having different concepts.

mkdocs/docs/reference/tasks/common.md
```diff
@@ -8,7 +8,7 @@ template: main.html
 
 ### Overview
 
-The `common/bash` task executes a bash script. The script can be written to use placeholders that are [dynamically substituted at runtime](../interpolation). For example, the `common/bash` task can be used as part of a finish action to promote the winning version at the end of an experiment.
+The `common/bash` task executes a bash script. The script can be written to use placeholders that are [dynamically substituted at runtime](../../tasks#dynamic-variable-substitution). For example, the `common/bash` task can be used as part of a finish action to promote the winning version at the end of an experiment.
```
sriumcp: Please use link in the following format `../tasks.md#dynamic-variable-substitution` (notice .md) and the relative path.

I am currently changing all links to be along the above lines (as documented by mkdocs).

sriumcp (changes requested):
minor change requested

mkdocs/docs/reference/tasks/common.md
kalantar: Done. Updated tasks.md as well since it was touched by this PR.

sriumcp (approved):
lgtm

PR #867, opened 2021-07-23
Replace bash with github action in gitops tutorial

kalantar:
Replace bash script with github action in istio/gitops tutorial.


huang195:
Looks good, please include doc changes before we merge

samples/istio/gitops/templates/experiment.yaml
```diff
@@ -13,36 +13,28 @@ spec:
...
+              }
+            }
+          headers:
+          - name: Accept
+            value: application/vnd.github.v3+json
+
+      - task: common/bash
+        with:
+          script: |
+            kubectl -n {{ .namespace }} apply -f https://raw.githubusercontent.com/iter8-tools/iter8/master/samples/istio/gitops/istio-vs.yaml
```
huang195: we don't want to hard code this URL

kalantar: I've replaced `iter8-tools` with `MY_ORG` so that it is replaced as part of the tutorial. There is no easy way to do otherwise without a script. Either script so can do git clone and local reference OR add to GH workflow but then need kubernetes credentials. This needs to be addressed in some way first.

.github/workflows/gitops-ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+        # give experiment a random name so CI triggers new experiment each time a new app version is available
+        sed "s|name: gitops-exp|name: gitops-exp-$RANDOM|" $ENV/templates/experiment.yaml > $ENV/experiment.yaml
+        cat $ENV/experiment.yaml
+        
+        # use a random color for a new experiment candidate
+        declare -a colors=("red" "orange" "blue" "green" "yellow" "violet" "brown")
+        color=${colors[$RANDOM % ${#colors[@]} ]}
+        sed "s|value: COLOR|value: \"${color}\"|" $ENV/templates/productpage-candidate.yaml > $ENV/productpage-candidate.yaml
+        cat $ENV/productpage-candidate.yaml
```
huang195: Do we want to keep these debugging message?

huang195:
Other than the minor nit, everything looks good. Feel free to merge or make changes.

huang195 (dismissed):

kalantar:
@huang195 this is will need to be re-approved after removing the cat statements.
@sriumcp I don't seem to have permission to merge code in this project.

huang195 (approved):

PR #915, opened 2021-07-30
fix your first experiment helm chart

sriumcp:


kalantar (dismissed):
I think the name "deploy" will cause us trouble later on. Perhaps should be renamed to "hello" or "hello-simple" or "first-experiment" or something so we know the purpose of the chart.

sriumcp:
> I think the name "deploy" will cause us trouble later on. Perhaps should be renamed to "hello" or "hello-simple" or "first-experiment" or something so we know the purpose of the chart.

Fixed now.

kalantar (approved):

PR #938, opened 2021-08-05
Openshift CI/CD non-functional doc

huang195:
@sriumcp @kalantar This is an attempt to make a non-functional doc for integrating iter8 into an existing CI/CD pipeline in Openshift per our discussion yesterday. We could make this more general than just  Openshift, but it looks like there isn't too much of a different Openshift vs. K8s.


sriumcp (approved):
lgtm

PR #944, opened 2021-08-06
Documentation, e2e test, and tutorial examples for conditions in tasks

### LINKED ISSUE #932: Enable conditional execution of tasks
Currently, there is no way to conditionally evaluate or skip the evaluation of a task within an action. Conditional evaluation of tasks can provide a cleaner alternative / enhancement to string interpolations currently used to dynamically change the behavior of tasks. 

I would like to define conditional evaluation within tasks as follows.
```yaml
- task: common/bash
    condition: WinnerFound()
    with: ... the usual stuff ...
- task: common/bash
    condition: CandidateWon()
    with: ... the usual stuff ...
- task: common/bash
    condition: not CandidateWon()
    with: ... the usual stuff ...
- task: common/bash
    condition: not WinnerFound()
    with: ... the usual stuff ...
- task: common/bash # run always
    with: ... the usual stuff ...
```

A task will be executed only if its condition is satisfied (if condition  is not specified, it will  always run).

Conditional evaluation can obviate the need for interpolation in tasks, which tend to be complicated. For example, in the `common/exec` task, without any versionRecommendedForPromotion (an optional field in experiment status), the task will fail; in the `common/bash` task, without any versionRecommendedForPromotion, the shell script will interpolate to garbage string and try to execute, probably resulting in failure or some other undesirable behavior which is opaque and hard to debug.

With the above condition, either the condition will succeed in which case the task can be executing using the with statement (without need for interpolation), or the condition will fail, meaning the task will simply not be executed.

There are a **lot** of conditions that can be imagined. I would start with `WinnerFound()`, and `CandidateWon()` are initial built conditions, with the option to use other free form expressions from the experiment object (like `.Status.NumIterations > 50`, just to provide a concocted example).

See design above. This is a non-breaking change to experiment CRD, and requires adding an optional `condition` field to task meta.

- [ ] Unit tested. E2E tested in tutorials.

- [ ] iter8.tools task documentation.

**Additional context**
We will leverage https://github.com/antonmedv/expr as part of the implementation. In addition to standard Go lang expression, there are some nice built in expressions that can be used in conditions with this library (like and / or / not , etc.).

sriumcp:
Fixes #932 

### DIFF
--- a/mkdocs/docs/reference/experiment.md
+++ b/mkdocs/docs/reference/experiment.md
@@ -119,6 +119,7 @@ Standard Kubernetes [meta.v1/ObjectMeta](https://kubernetes.io/docs/reference/ge
 | Field name | Field type | Description | Required |
 | ----- | ---- | ----------- | -------- |
 | task | string | Name of the task. Task names express both the library and the task within the library in the format 'library/task' . | Yes |
+| condition | string | Execute the task only if the specified condition is satisfied. Two built-in conditions are supported namely, `WinnerFound()` and `CandidateWon()`. They can be combined with predicates like `not`: `not WinnerFound()` and `not CandidateWon()` are also valid conditions.[^2] | No |
 | with | map[string][apiextensionsv1.JSON](https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1#JSON) | Inputs to the task. | No |
 
 ### Criteria
@@ -304,3 +305,4 @@ Standard Kubernetes [meta.v1/ObjectMeta](https://kubernetes.io/docs/reference/ge
 | value | string | Value of a variable. | Yes |
 
 [^1]: `A/B/n` experiments involve more than one candidate. Their description is coming soon.
+[^1]: This feature is built using the [`expr` Go expression library](https://github.com/antonmedv/expr). The [Experiment Go struct](https://pkg.go.dev/github.com/iter8-tools/etc3@v0.1.30/api/v2alpha2#Experiment) is used as the environment for expression evaluation. Hence, `.Status.Analysis != nil` is also a valid condition.
--- a/install/core/iter8-controller/kustomization.yaml
+++ b/install/core/iter8-controller/kustomization.yaml
@@ -1,12 +1,12 @@
 resources:
 - etc3.yaml
 # to change Iter8 CRDs version, edit the next line
-- github.com/iter8-tools/etc3/config/crd/?ref=v0.1.29
+- github.com/iter8-tools/etc3/config/crd/?ref=v0.1.30
 
 images:
 - name: iter8/etc3:latest
   # to change etc3 version, edit the next line
-  newTag: 0.1.29
+  newTag: 0.1.30
 
 # even if you do not change the CRD, 
 # keep the two version numbers in sync to keep them the same
--- a/install/core/iter8-handler/handler.yaml
+++ b/install/core/iter8-handler/handler.yaml
@@ -9,7 +9,7 @@ spec:
       containers:
       - name: iter8-handler
         # to change handler version, edit the line below
-        image: iter8/handler:0.1.24
+        image: iter8/handler:0.1.25
         command: ["handler"]
         args: ["run", "-a", "$(ACTION)"]
         env:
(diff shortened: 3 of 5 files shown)


mkdocs/docs/reference/experiment.md
```diff
@@ -304,3 +305,4 @@ Standard Kubernetes [meta.v1/ObjectMeta](https://kubernetes.io/docs/reference/ge
 | value | string | Value of a variable. | Yes |
 
 [^1]: `A/B/n` experiments involve more than one candidate. Their description is coming soon.
+[^1]: This feature is built using the [`expr` Go expression library](https://github.com/antonmedv/expr). The [Experiment Go struct](https://pkg.go.dev/github.com/iter8-tools/etc3@v0.1.30/api/v2alpha2#Experiment) is used as the environment for expression evaluation. Hence, `.Status.Analysis != nil` is also a valid condition.
```
kalantar, line 308: There is a mismatch in how we evaluate placeholders using the experiment. There we use `.this` to refer to the experiment.  So, `.this.Status.Analysis`. Can something similar be done here as well?

sriumcp: I think `.Status.Analysis` is preferable to `.this.Status.Analysis` since it is shorter. Perhaps we can work towards getting rid of `.this` in the variable interpolation as well?

It sort of makes sense that the default source of data for interpolation is the experiment struct.

sriumcp: Besides note that `CandidateWon()`, `not CandidateWon()`, `WinnerFound()`, and `not WinnerFound()` are enough to cover almost all the use-cases we have currently documented within the tutorials.

I think `.Status.Analysis....` itself is less preferable to meaningful builtins such as the above.

sriumcp: With the new task specific versionInfo / expression evaluation libraries, we have more options opening up ... we can eventually extend tasks to do the following.
```yaml
- name: lib/task
  condition: <your condition here>
  for: WinnerOrBaseline() # Winner(),  Baseline(), Candidate(), not Winner() are some other options
  with:
    versionInfo:
    - variables:
  ...
```

This gets rid of version-specific variable interpolation altogether and simply selects the version(s) **for** which the task will run. It will use the selected version's versionInfo.

sriumcp: Actually, we can do one better ... according to [Golang template reference](https://pkg.go.dev/text/template#hdr-Arguments) ... 

```yaml
- name: common/shell
  condition: WinnerFound()
  with:
   # .Winner changes the scope where the template variables are evaluated
    script: kubectl apply -f {{ .Winner | .manifest }}
    versionInfo:
    - variables: # for baseline
      - name: manifest
        value: https://baseline.manifest.yaml
    - variables: # for candidate
      - name: manifest
        value: https://candidate.manifest.yaml
```

```yaml
- name: common/shell
  with:
   # .WinnerOrBaseline changes the scope where the template variables are evaluated
    script: kubectl apply -f {{ .WinnerOrBaseline | .manifest }}
    versionInfo:
    - variables: # for baseline
      - name: manifest
        value: https://baseline.manifest.yaml
    - variables: # for candidate
      - name: manifest
        value: https://candidate.manifest.yaml
```

sriumcp: It would be nice to get rid of version recommended for promotion altogether. Because, users need to understand **two** separate concepts, winner, as well as version recommended for promotion instead of only one. The above accomplishes this simplification.

Add to the fact that versionRecommendedForPromotion has issues (it is optional; not always set), and `common/bash` has undefined behavior when it is not set, it is actually a recipe for subtle failures that are difficult to debug.

sriumcp: Issue created: https://github.com/iter8-tools/iter8/issues/945

huang195 (approved):
/lgtm

mkdocs/docs/reference/experiment.md
kalantar: > I think `.Status.Analysis` is preferable to `.this.Status.Analysis` since it is shorter. Perhaps we can work towards getting rid of `.this` in the variable interpolation as well?
> 
> It sort of makes sense that the default source of data for interpolation is the experiment struct.

I am a bit ambivalent. I chose `.this` because I wanted to avoid possible conflicts with user variable names. While `.metadata` and `.spec` do seem unlikely, I thought `.status` seemed as likely as any other variable name.  `.this` seemed safe.

Agree that a couple of well defined functions will go a long way to making this irrelevant. I think we still need to learn which variables are most likely to be read.

lgtm

Issue #935, opened 2021-08-05
Chaos injection during experiments


sriumcp:
**Is your feature request related to a **problem**? Please describe the **problem**.**
[Litmus](https://litmuschaos.io) provides excellent ways of injecting chaos for K8s applications and chaos test them. Currently, there are no examples that demonstrate chaos testing in conjunction with SLO validation.

**Describe the feature/solution you'd like**
Create an end-to-end example that demonstrates how chaos injection and SLO validation can be performed together.

- [ ] Modify the Helm templated used in [Iter8's your first experiment tutorial](https://iter8.tools/0.7/getting-started/first-experiment/)
- [ ] Modified template needs to parameterize replica count (defaulted to 1)
- [ ] Modified template needs to include chaos resource
- [ ] Modified template needs to include RBACs needed by chaos to check for Iter8 readiness and vice-versa (check creation of Chaos results resource; check creation of Iter8 experiment resource -- must be initialized)
- [ ] At the end of the experiment, `iter8ctl assert WinnerFound` (initially this will fail). 
- [ ] At the end of the experiment, inspect chaos verdict (initially this will fail). 
- [ ] Rollback candidate (this deletes the dark launched version; it will not be promoted)

This completes the initial scenario.
***
- [ ] Run-run the above scenario with sufficiently high replica count
- [ ] Do the same steps as above
- [ ] At the end of the experiment, both `iter8ctl` and chaos verdicts will be success.
- [ ] Promote candidate (this deletes the old baseline  version and promotes the new candidate as the baseline version)
***

**Will this feature/solution bring new benefits for Iter8 users? What are they?**
How will this feature/solution add value to Iter8 users.

**Will this feature/solution bring new benefits for Iter8 developers? What are they?**
How will this feature/solution add value to Iter8 developers.

**Does this issue require a design doc/discussion? If there is a link to the design document/discussions, please provide it below.**
You can use this issue thread, or a discussion thread in this repo, for design discussions. Please provide the appropriate links. You can use the issue/discussion to also discuss and describe alternatives.

**How will this feature be tested?**
- [ ] A checklist of behaviors that will be tested as part of this feature implementation. 

**How will this feature be documented?**
- [ ] Most new features require end-user documentation at https://iter8.tools. Describe the specific documentation that will be added.

**Additional context**

sriumcp:
Under `$ITER8/samples`, lets create a `chaos` folder, and put all the source code, include Helm charts within the `chaos` folder.

PR #956, opened 2021-08-10
feat(chaos): Adding litmus chaos helm template

ispeakc0de:
Signed-off-by: shubham chaudhary <shubham@chaosnative.com>

 Adding litmus chaos helm template. Which will install the following items:
- RBAC for all the resources
- litmus CRDs
- chaos-operator(control plane for litmus) [in namespace and  cluster scoped both]
- pod-delete experiment CR
- pod-delete engine CR

ref: #935

### DIFF
--- a/samples/chaos/charts/Chart.yaml
+++ b/samples/chaos/charts/Chart.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+appVersion: "1.13.8"
+description: A Helm chart to install litmus infra components on Kubernetes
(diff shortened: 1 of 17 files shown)


sriumcp (approved):
@ispeakc0de I am approving and merging this PR so that we can move on to the next step (creating a tutorial out of this).

However, I feel we will want to revisit this code so that installation of chaos engine can be done as part of the setup step (and can use the chaos documentation directly).

lgtm.

PR #966, opened 2021-08-16
CICD tutorial on Openshift

huang195:
This PR includes the resource files used for the Openshift CI/CD demo. These are going to be "hidden" until a blog that makes use of these is ready. 

@sriumcp I don't think a review is needed here until the blog ready, but please take a look.


sriumcp (approved):
lgtm. @huang195 Please feel free to merge. I am curious about the other OpenShift GitOps tutorial currently present in the documentation here ... what would it fate be? https://iter8.tools/0.7/tutorials/openshift/openshift/ I am guessing this stays?

PR #985, opened 2021-08-31
GitHub Workflow to change install version

Alan-Cha:
Replace #981


mkdocs/mkdocs.yml
```diff
@@ -185,9 +199,4 @@ nav:
     - Analytics: contributing/analytics.md
     - Tasks: contributing/tasks.md
   - Roadmap: roadmap.md
-  - News: news.md
-
-# Google Analytics
-google_analytics:
```
kalantar, line 191: Did we intend to remove google_analytics directive?

.github/workflows/ind.yaml
```diff
@@ -29,15 +24,11 @@ jobs:
...
         fi
-    - name: Get owner and repo
-      run: |
-        ownerrepo=${{ github.repository }}
-        owner=$(echo $ownerrepo | cut -f1 -d/)
-        if [[ "$owner" == "iter8-tools" ]]; then
-          owner=iter8
-        fi
-        echo "OWNER=$owner" >> $GITHUB_ENV
-        echo "REPO=ind" >> $GITHUB_ENV
```
kalantar, line 40: aren't OWNER and REPO still needed? I still see them in the `tags` argument to build-push-action

.github/workflows/change-install-version.yaml
```diff
@@ -0,0 +1,31 @@
...
+        elif [[ "${{ github.ref }}" == *"master" ]]; then
+          echo "VERSION=latest" >> $GITHUB_ENV
+        fi
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: pip3 install -r ./.github/workflows/requirements.txt
+    - name: Edit install version
+      run: python3 ./.github/workflows/change-install-version.py ${{ env.VERSION }}
```
kalantar, line 27: I'm not understanding how this updates `mkdocs/docs/getting-started/install.md`. I see the py script updates only `mkdocs.yml`.

.github/workflows/change-install-version.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - uses: EndBug/add-and-commit@v7
```
kalantar, line 28: Do we want to just commit? Or do we want to create PR as well?

kalantar (commented):
I have some questions...

mkdocs/mkdocs.yml
Alan-Cha: This method of adding Google Analytics is deprecated. I've added an updated version.

mkdocs/mkdocs.yml
```diff
@@ -53,9 +53,20 @@ plugins:
...
+  - macros:
+      # Replace {{ variable }} syntax with [[ variable ]] to avoid conflicts
+      j2_block_start_string: '[[%'
+      j2_block_end_string: '%]]'
+      j2_variable_start_string: '[['
+      j2_variable_end_string: ']]'
         
 # Customization
 extra:
+  # Google Analytics
```
Alan-Cha, line 65: @kalantar This is the new method of adding Google Analytics. You can read more about it [here](https://github.com/iter8-tools/iter8/pull/981#discussion_r693106545).

.github/workflows/change-install-version.yaml
Alan-Cha: I've replaced the install version with a variable. Now we can dynamically update the install version by modifying `mkdocs.yml`.

mkdocs/docs/getting-started/install.md
```diff
@@ -9,7 +9,7 @@ hide:
 Install Iter8 in your Kubernetes cluster as follows. This step requires [Kustomize v3+](https://kubectl.docs.kubernetes.io/installation/kustomize/).
 
 ```shell
-export TAG=v0.7.19
+export TAG=[[ iter8.install_version ]]
```
Alan-Cha, line 12: @kalantar This is the install version variable.

mkdocs/mkdocs.yml
```diff
@@ -64,7 +75,10 @@ extra:
       link: https://github.com/iter8-tools/iter8
     - icon: fontawesome/brands/slack
       link: https://join.slack.com/t/iter8-tools/shared_invite/zt-awl2se8i-L0pZCpuHntpPejxzLicbmw
-
+  
+  iter8:
+    install_version: v0.7.19
```
Alan-Cha, line 80: @kalantar This is where we can change the install version in the `mkdocs.yml`.

.github/workflows/ind.yaml
Alan-Cha: I think this may have been a mistake.

.github/workflows/change-install-version.yaml
Alan-Cha: I think committing is less abrasive. If we were to go the other route, then we would need to wait for a workflow and accept a PR after every new release. I think we go with commits for now and we can change it in the future if the requirements are different.

kalantar: This isn't just commit; it is push as well. Any opinions @sriumcp and @huang195 ?

mkdocs/mkdocs.yml
kalantar: Need to be updated already.

Alan-Cha:
See #995

PR #987, opened 2021-09-08
changes for merged repos

kalantar:
Modifications to iter8 necessary to support merged etc3/iter8ctl/handler repos.


kalantar:
@sriumcp I don't think we envisioned this to be merged with master. Rather it was meant as proof of concept; that the etc3 changes would work once they were merged. This PR will need to be modified to refer to the tag once we merge and create.

Could you have a look at this in terms of whether or not there is anything missing or if this is sufficient motivation to move forward with the etc3 changes.

sriumcp:
@kalantar All checks have passed which IMO is sufficient motivation to move forward with `etc3` changes.

kalantar:
@sriumcp one thing I noticed is that we need to update the etc3 version in each test file. There are more places to make sure it is right. Is there a way we can simplify this?  Perhaps a manually executed wf to create PR or something?

sriumcp:
You mean due to `go get iter8ctl`? 

I think we may be able to drop the version specification altogether and simply go with the latest. Please create an issue in anycase. I feel we don't need yet another workflow (I would like to get rid of a few of them if possible).

.github/workflows/e2e-tests.yaml
```diff
@@ -33,13 +39,8 @@ jobs:
...
         $ITER8/samples/setup/iter8-setup.sh
 
-    - name: Install Go
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1.16
-    
     - name: Get iter8ctl
-      run: sudo GO111MODULE=on GOBIN=/usr/local/bin go get github.com/iter8-tools/iter8ctl@v0.1.6
+      run: go install github.com/iter8-tools/etc3/iter8ctl@latest
```
sriumcp, line 43: I believe we should drop @latest (we want the latest *semantically versioned* release).

`go install github.com/iter8-tools/etc3/iter8ctl` is probably the right thing to do to achieve the above.

.github/workflows/knquickstart.yaml
```diff
@@ -42,6 +42,12 @@ jobs:
 
     steps:
 
+    # install Go locally
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
```
sriumcp, line 49: go-version requirements seem different. Is this intentional?

.github/workflows/knquickstart.yaml
```diff
@@ -59,17 +65,11 @@ jobs:
...
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1.16
-    
     # get iter8ctl;
     # among other things, this Iter8 CLI can help verify that 
     # the experiment completed successfully.
     - name: Get iter8ctl
-      run: sudo GO111MODULE=on GOBIN=/usr/local/bin go get github.com/iter8-tools/iter8ctl@v0.1.6
+      run: go install github.com/iter8-tools/etc3/iter8ctl@latest
```
sriumcp, line 72: See comment on `latest` above.

install/docker/Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN chmod +x ./kustomize
 RUN mv ./kustomize /bin/kustomize
 
 # get iter8ctl
-RUN GO111MODULE=on GOBIN=/usr/local/bin go get github.com/iter8-tools/iter8ctl@v0.1.4
+RUN GO111MODULE=on GOBIN=/usr/local/bin go install github.com/iter8-tools/etc3/iter8ctl@latest
```
sriumcp: Please see comment on `go install` of iter8ctl without latest above.

mkdocs/docs/getting-started/install.md
```diff
@@ -19,7 +19,7 @@ kubectl wait --for=condition=Ready pods --all -n iter8-system
 ## Get `iter8ctl`
 Get `iter8ctl` CLI on your local machine as follows. This step requires [Go 1.16+](https://golang.org/doc/install).
 ```shell
-GO111MODULE=on GOBIN=/usr/local/bin go get github.com/iter8-tools/iter8ctl@v0.1.6
+go install github.com/iter8-tools/etc3/iter8ctl@latest
```
sriumcp, line 22: Please see comment on `go install` of iter8ctl without `latest` above.

sriumcp (changes requested):
Please see comments on `iter8ctl` without `latest` tag.

.github/workflows/e2e-tests.yaml
kalantar: `@latest` seems to be required: 

```bash
$ go install github.com/iter8-tools/etc3/iter8ctl
go install: version is required when current directory is not in a module
	Try 'go install github.com/iter8-tools/etc3/iter8ctl@latest' to install the latest version
```

Perhaps by moving etc3 into a subdirectory (possible after v0.8 integration) and creating modules in each we can solve this. But for now, no.

.github/workflows/knquickstart.yaml
kalantar: Yes. It was my thought that we can get more coverage by testing different versions of go in e2etests and knquickstart.  I can make them the same if you think this not worth doing.  Thoughts?

.github/workflows/e2e-tests.yaml
sriumcp: I am able to install without `@latest`. I use the full `GO111MODULE=on ... ` command locally.

kalantar: This seemed to work when I tried locally. But I get an error when running the workflow. See https://github.com/iter8-tools/iter8/pull/987/checks?check_run_id=3666987855

sriumcp: I am getting a different error ... https://github.com/iter8-tools/iter8/runs/3666504483?check_suite_focus=true

I thought this error was fixed ... anyway, it does seem to be getting the right install version for me though!

kalantar: I fixed this by moving the install of go earlier in the workflow.

sriumcp: @kalantar https://pkg.go.dev/cmd/go -- use of `@latest` seems to be the right way to do this (latest does not refer to master branch of the Git repo; it refers to the latest semantically ordered version, which is the interpretation that makes sense and works for us).

So all good with the use of `@latest`. This is what we need. Sorry about creating a confusion in this regard.

sriumcp (approved):
lgtm

.github/workflows/e2e-tests.yaml
kalantar: I've reverted to @latest.

PR #989, opened 2021-09-08
First exp

### LINKED ISSUE #988: Create Helm template for your first experiment without templating the app
Currently, your-first-experiment uses a Helm template that templates not only the experiment but also the application. This makes it less obvious how to use the experiment for users who do not use Helm to deploy their application.

Template the experiment. Not the application. Also, make it clear how users can use the experiment with *their* application.

Make the tutorial more readily applicable to the users' application. Also make it applicable to users who do not use Helm to deploy their application (they still will use Helm to deploy the Iter8 experiment).

sriumcp:
Fixes #988 

Replace `Install iter8ctl` with `Get iter8ctl` everywhere. Minor change -- but we tend to use the verb `get` for local clients (example, get helm), and `install` for installing in a cluster (install iter8). This replacement makes the verbs consistent.


.github/workflows/knquickstart.yaml
```diff
@@ -65,10 +65,10 @@ jobs:
       with:
         go-version: 1.16
     
-    # install iter8ctl;
+    # get iter8ctl;
     # among other things, this Iter8 CLI can help verify that 
     # the experiment completed successfully.
-    - name: Install iter8ctl
+    - name: Get iter8ctl
       run: sudo GO111MODULE=on GOBIN=/usr/local/bin go get github.com/iter8-tools/iter8ctl@v0.1.5
```
kalantar: Update to v0.1.6 everywhere.

kalantar (changes requested):

kalantar (approved):

PR #990, opened 2021-09-08
Gitops

### LINKED ISSUE #943: Drop `Helmex-update` in favor of `run` and make it gitops friendly
Git clone and commit are now bundled together into Helmex-update. This makes it impossible to use those functions in other  settings (like non-Helm GitOps).

Drop helmex-update and add support for a proper use of secrets for git cloning in `common/bash`. This will enable us to rewrite the current experiment using `helmex-update` using the following three tasks, and also other non-helm based GitOps.
1. Bash task to git clone using a token (so that 3 below becomes possible)
2. Bash task to edit yamls locally
3. Bash task to  commit / push

Enables non-Helm based GitOps.

Yes, please see above.

- [ ] Unit tested. E2e tested as part of tutorials. Manually tested as well.

- [ ] Iter8.tools end user documentation.

sriumcp:
Fixes #943 

This is the GitOps tutorial we need. It uses Helm templated experiment and also uses the new `run` construct.


sriumcp:
To be reopened.

PR #992, opened 2021-09-10
gitops

sriumcp:
- gitops experiment
- fixes to first experiment before gitops
- stable version
- candidate service and deployment
- candidate
- updated gitops experiment tutorial
- slo gitops chart
- gitops slo experiment
- gitops slo experiment with literal braces
- using yq instead of kubectl


sriumcp:
Will be reopened

PR #1004, opened 2021-09-21
Fix detached head issue in post release changes workflow

Alan-Cha:


.github/workflows/post-release-changes.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 0 # fetch all commits/branches
+        ref: refs/heads/master # checkout master branch instead of tagged release
```
Alan-Cha, line 19: The problem was that because we configured the workflow to run on published releases, the checkout action would checkout a tagged release, which would have a detached head, meaning we could commit to it.

I wanted to properly get the branch from the tagged release but I was having trouble doing so. I think hard coding it here will work fine though.

PR #1020, opened 2021-09-28
openshift gitops

huang195:
This PR contains manifests needed to demonstrate an openshift gitops scenario using CodeReady Container. 

This is a headless PR, meaning there's no documentation / blogs referring to this, yet.


.github/workflows/gitops-ci-openshift.yaml
```diff
@@ -0,0 +1,31 @@
+name: Simulate CI pipeline on Openshift
```
sriumcp, line 1: Many of our tutorials involve GH action workflows, which is great. It would be nice if we can split up all the GH action workflows we have under the .github/workflows folder into subfolders so that it is clear what belongs to Iter8's own CI and what belongs to Iter8 tutorials.

For starters, it would be nice to simply create .github/workflows/tutorials and move this under that folder.

.github/workflows/gitops-finish-openshift.yaml
```diff
@@ -0,0 +1,53 @@
+name: Update baseline version on Openshift
```
sriumcp, line 1: Same comment as the other ci workflow.

samples/openshift/gitops/details.yaml
```diff
@@ -0,0 +1,68 @@
+# Copyright 2017 Istio Authors
```
sriumcp, line 1: I feel the bookinfo app is an overkill for us to demonstrate Iter8 related concepts.

In addition to complexity of GitOps / CI / CD concepts, the app is adding needless complexity through multiple services, deployments and gateways.

The hello world application at that is used in the getting started tutorial, and also Iter8's other GitOps tutorial for that matter is a whole lot simpler. In light of the [simplicity of this workflow](https://iter8.tools/0.7/tutorials/deployments/slo-validation-gitops/), do you think the OpenShift GitOps tutorial can be further simplified? In particular, using hello world instead of bookinfo? 

Same is true for all of our Istio tutorials. While bookinfo is excellent for demonstrating Istio, I'm not sure why we need all that complexity for Iter8.

samples/openshift/gitops/templates/experiment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - task: metrics/collect
+        with:
+          time: "5s"
+          versions:
+          - name: productpage-candidate
+            url: "http://productpage-candidate.default.svc.cluster.local:9080"
+            qps: 8
+      # when the experiment completes, promote the winning version in the Env repo
+      finish:
+      - task: notification/http
```
sriumcp, line 22: This is fantastic. I like it that you are using `notification/http`.

sriumcp (commented):
Initial review comments.

sriumcp:
I wanted to add yet another comment to the review, but forgot :-) 

Would you also consider Helm templating the experiment for reusability? An example of a Helm templated GitOps experiment is [here](https://iter8.tools/0.7/tutorials/deployments/slo-validation-gitops/). In particular, if you expand the next steps, you can see the simplicity of reuse, thanks to Helm.

.github/workflows/gitops-ci-openshift.yaml
huang195: Make sense. Let's gradually make this change, starting from this PR.

samples/openshift/gitops/details.yaml
huang195: I agree with this comment. However, I want to show some screenshots of the app itself on the blog article that I'm writing, so it's nice to have a prettier app that looks more like a real app than the helloworld app. Also, if you look at `...samples/openshift/gitops`, there are only a few yaml files, and they're all encapsulated by `argocd-app.yaml`, so it's already pretty simple.

huang195:
> Would you also consider Helm templating the experiment for reusability? An example of a Helm templated GitOps experiment is here. In particular, if you expand the next steps, you can see the simplicity of reuse, thanks to Helm.

This scenario is primarily driven by the blog article I'm currently writing, and Helm doesn't seem to fit into the current flow but adds another dependency. As the plan is to have followup articles after the initial one is published, I think Helm and Openshift Pipeline are topics that would fit better in these subsequent ones.

.github/workflows/gitops-ci-openshift.yaml
huang195: ok, I tried creating subdirectories within .github/workflows/ and I was not able to manually trigger the actions anymore. This stackoverflow article explained that subdirectories are not currently supported: https://stackoverflow.com/questions/62288443/where-should-you-put-the-git-hub-workflow-directory-for-actions-in-a-full-stack. As a result, I'm reverting the change until Github supports subdirectories for Github Actions.

sriumcp (approved):
lgtm

PR #1030, opened 2021-10-13
some istio experiment templates

kalantar:

### DIFF
--- a/iter8hub/istio/slo-validation-1-builtin/Chart.yaml
+++ b/iter8hub/istio/slo-validation-1-builtin/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v2
+name: slo
+description: A Helm chart for Iter8 SLO validation experiments
+type: application
+version: 0.1.0
--- a/iter8hub/istio/slo-validation-1-builtin/README.md
+++ b/iter8hub/istio/slo-validation-1-builtin/README.md
@@ -1,0 +1,32 @@
+# Use Case
+
+SLO validation of a single version of an application deployed to Istio.
+Test with built-in metrics.
+
+## Comments
+
+This just explores the template of the experiment; no effort to convert to a configmap or to incluce RBAC rules is included.
+
+There is nothing special due to Istio; the same helm chart that is used for any other SLO validation with built-in metrics can be used.
+
+Note, however, that it is slightly more general than other SLO valiation charts we have developed in the past. It enables use of a header in the request.
+
+```yaml
+          {{- if .Values.version.headers }}
+          headers:
+{{ toYaml .Values.version.headers | indent 12 }}
+          {{ end }}
+```
+
+## Sample Use
+
+```bash
+helm upgrade -n $NAMESPACE slo-experiment $ITER8/samples/istio/slo-validation \
+  --set limitMeanLatency=100.0 \
+  --set limitErrorRate=0.0 \
+  --set limit95thPercentileLatency=200.0 \
+  --set version.name=hello \
+  --set version.URL=http://istio-ingressgateway.istio-system.cluster.local:80/hello \
+  --set version.headers.host=hello.example.com \
+  --install 
+```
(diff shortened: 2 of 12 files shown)


sriumcp (changes requested):
I have a few requests for the builtin as well as the prometheus chart.

Let me start with the built-in chart.

1. Enabling headers as part of the chart install is an excellent idea.
2. The latest SLO validation chart used in the payload tutorial is [here](https://iter8.tools/0.7/tutorials/deployments/slo-validation-payload/). It can be used to send payloads as well. We can incorporate payload related fields in this chart too. 
3. `version.headers.host` has 8 extra characters. `headers.host` would suffice.
4. Please replace `version.name` with a non-nested `appName` value and give it a default value so that users don't have to specify this as part of the chart install. The documented example should favor this pattern (since it makes zero difference to the experiment whether the app is called hello, my-app, your-app or whatever).
5. Incorporating a (single, fixed) uniqueId suffix as part of every resource installed by the Helm chart is a good idea. For starters, it would ensure that subsequent Helm upgrades of the experiment work correctly. Please see the Helm charts in the first five tutorials, all of which do this now.
6. Given that the goal is indeed getting to 0.8, could you please add the configmap and RBAC resources in this Helm chart? Doing so, even for only a single experiment (with builtin metrics with no other frills) is already good first step in the 0.8 design.

kalantar:
> 3. `version.headers.host` has 8 extra characters. `headers.host` would suffice.
> 4. Please replace `version.name` with a non-nested `appName` value and give it a default value so that users don't have to specify this as part of the chart install. The documented example should favor this pattern (since it makes zero difference to the experiment whether the app is called hello, my-app, your-app or whatever)

I was trying to explore 2 questions with this chart the other others.

First, can we write a chart that works for 1 version or 2 versions (probably)? Can I write a chart that works for generated load and prometheus or non-generated load and prometheus (probably)? Can I write one for built-in and prometheus (probably not)? 

Second, I asked myself -- if I have to change a chart between the above scenarios, how should we express it so that there are as few -- and as obvious -- differences between them? 

My thinking is that we want to create charts that are as useful across scenarios as possible, that require a new chart in well understood and easily articulated circumstances and which enable chart creators/uses to easily understand the options for one based on what is provided in another.

For this reason, I introduced `version`. If we move to 2 version use cases, this just becomes `versionA` and `versionB` or `versions` (and a list). This is an easily understood change. And the variables that need to be replicated are easily understood; hence`version.URL` and `version.headers`.

By the same token, I was thinking that I would prefer to capture most of the load generation variables using a group such as `generateLoad`. This adds characters but makes clear which are used for load generation. The concept of Iter8 generated load is very general and applies in many charts that we create. Would be nice to logically group so this is obvious and it is easy to cut and paste into new charts.

kalantar:
> 6. Given that the goal is indeed getting to 0.8, could you please add the configmap and RBAC resources in this Helm chart? Doing so, even for only a single experiment (with builtin metrics with no other frills) is already good first step in the 0.8 design.

I added [some thoughts on how to express RBAC rules](https://docs.google.com/document/d/1NY18q-TpQdLSppBlOO44ejLxrvJczYe1zGDEpCmSeE4/edit#bookmark=id.y6w5ha53wb3x) to our design doc. Perhaps we could consider in this context?

sriumcp:
Today, we have separate experiments documented in our tutorials for each use-case. Without manual edits, the experiments cannot be re-used across different applications. Without manual edits, the experiments cannot be re-used across different releases of the same application by the same user. This is the limitation of the non-Helm tutorials today in iter8.tools.
***
The initial attempt should be to create Helm charts, one per use-case, which can be reused across releases of the app, and across different apps. Once we have some feedback from adopters of these Helm charts, we can certainly think about how to create cross-cutting ones that work across multiple use-cases. But I think trying to create a universal chart right off the bat is  pre-mature optimization, and the resultant complexity would hurt the community adoption of these charts, not to mention making the charts complex and hard to maintain and update in the future.
***
Helm specifically discourages nested values. Here's what Helm has to say on this matter. Verbatim.
```
In most cases, flat should be favored over nested. The reason for this is that it is simpler for template developers and users.
```

***
If we want a single chart that works across use-cases, where do we draw the line? Why stop with builtin metrics & Istio/Prometheus? Why not Linkerd/prom also? What about the GitOps Helm examples and GitHub action examples we've already documented?
***

PR #1036, opened 2021-10-19
Add GitHub buttons

Alan-Cha:


Alan-Cha:
I think my commits should be properly verified now! @sriumcp Let me know!

sriumcp (approved):
lgtm

sriumcp:
@Alan-Cha Still shows up as unverified. But given how important this fix is, I merged anyway.

PR #1044, opened 2021-11-04
update broken references after files were moved

huang195:
Fixes the following problems:
- Some references in yaml files were broken when other files moved around
- versionInfo variable substitution doesn't work anymore in notification/http task, so using WinnerFound() in experiment.yaml

### DIFF
--- a/samples/openshift/full-cicd/templates/experiment.yaml
+++ b/samples/openshift/full-cicd/templates/experiment.yaml
@@ -6,9 +6,8 @@ spec:
   # target identifies the service under experimentation using its fully qualified name
   target: default/productpage
   strategy:
-    # this experiment will perform an A/B test
-    testingPattern: Canary
-    deploymentPattern: FixedSplit
+    # this experiment will perform a conformance test
+    testingPattern: Conformance
     actions:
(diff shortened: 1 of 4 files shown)


samples/openshift/full-cicd/templates/experiment.yaml
```diff
@@ -21,28 +20,43 @@ spec:
             url: "http://iter8-app-candidate.default.svc.cluster.local:8000"
       # when the experiment completes, promote the winning version in the Env repo
       finish:
-      - task: notification/http
+      - if: WinnerFound()
+        task: notification/http
         with:
-          url: https://api.github.com/repos/MY_ORG/iter8/actions/workflows/gitops-finish.yaml/dispatches
+          url: https://api.github.com/repos/huang195/iter8/actions/workflows/gitops-finish-openshift.yaml/dispatches
```
kalantar: Do you want to change from MY_ORG?

huang195: Fixed

sriumcp (approved):
lgtm

PR #1054, opened 2021-11-19
V0.8

sriumcp:
Refactored code.

Enables reuse of Gen, assert, and run functionality, and extensions to Build and Builtin template functionality.


base/collect.go
```diff
@@ -37,6 +37,8 @@ type collectInputs struct {
 	Duration *string `json:"duration" yaml:"duration"`
 	// QPS is the number of queries per second sent to each version. Default is 8.0.
 	QPS *float32 `json:"qps" yaml:"qps"`
+	// Connections is the number of number of parallel connections used to send load. Default is 4.
```
sriumcp, line 40: In response to comments from to OpenShift Serverless team.

cmd/assert.go
```diff
@@ -64,81 +64,92 @@ var assertCmd = &cobra.Command{
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
 		// build experiment
-		exp := &experiment{
+		exp := &Experiment{
 			Experiment: &base.Experiment{},
 		}
 		log.Logger.Trace("build started")
-		exp, err := build(true)
+		// replace FileExpIO with ClusterExpIO to build from cluster
```
sriumcp, line 71: Here is where FileIO can be replaced with ClusterIO

cmd/assert.go
```diff
@@ -64,81 +64,92 @@ var assertCmd = &cobra.Command{
...
-						os.Exit(1)
-					}
-					iv := exp.SLOsBy(version)
-					allGood = allGood && iv
-					if iv {
-						log.Logger.Info(version, " satisfies objectives")
-					} else {
-						log.Logger.Info(version, " does not satisfy objectives")
-					}
+		allGood, err := exp.Assert(conds, timeout)
```
sriumcp, line 80: Reuse assert

cmd/experiment.go
```diff
@@ -72,40 +78,62 @@ func build(withResult bool) (*experiment, error) {
...
 
-// read experiment spec from file
-func readSpec() ([]base.TaskSpec, error) {
-	yamlFile, err := ioutil.ReadFile(experimentSpecPath)
-	if err != nil {
-		log.Logger.WithStackTrace(err.Error()).Error("unable to read experiment spec")
-		return nil, errors.New("unable to read experiment spec")
-	}
+//FileExpIO enables reading and writing through files
+type FileExpIO struct{}
```
sriumcp, line 86:
```go
type ClusterExpIO struct {
  id: string // id of the experiment
}
```

cmd/experiment.go
```diff
@@ -72,40 +78,62 @@ func build(withResult bool) (*experiment, error) {
...
+
+// SpecFromBytes reads experiment spec from bytes
+func SpecFromBytes(b []byte) ([]base.TaskSpec, error) {
```
sriumcp, line 89: Reuse for ClusterIO

cmd/experiment.go
```diff
@@ -72,40 +78,62 @@ func build(withResult bool) (*experiment, error) {
...
+	err := yaml.Unmarshal(b, &e)
 	if err != nil {
 		log.Logger.WithStackTrace(err.Error()).Error("unable to unmarshal experiment spec")
 		return nil, err
 	}
 	return e, err
 }
 
+// ResultFromBytes reads experiment result from bytes
+func ResultFromBytes(b []byte) (*base.ExperimentResult, error) {
```
sriumcp, line 100: Reuse for ClusterIO

cmd/experiment.go
```diff
@@ -72,40 +78,62 @@ func build(withResult bool) (*experiment, error) {
...
+		log.Logger.WithStackTrace(err.Error()).Error("unable to read experiment spec")
+		return nil, errors.New("unable to read experiment spec")
+	}
+	return SpecFromBytes(b)
+}
+
 // read experiment result from file
-func readResult() (*base.ExperimentResult, error) {
-	yamlFile, err := ioutil.ReadFile(experimentResultPath)
+func (f *FileExpIO) readResult() (*base.ExperimentResult, error) {
```
sriumcp, line 121: rewrite for ClusterIO

cmd/experiment.go
```diff
@@ -72,40 +78,62 @@ func build(withResult bool) (*experiment, error) {
...
 		log.Logger.WithStackTrace(err.Error()).Error("unable to read experiment result")
 		return nil, errors.New("unable to read experiment result")
 	}
-	e := &base.ExperimentResult{}
-	err = yaml.Unmarshal(yamlFile, e)
+	return ResultFromBytes(b)
+}
+
+// write experiment result to file
+func (f *FileExpIO) writeResult(r *Experiment) error {
```
sriumcp, line 131: rewrite for ClusterIO

cmd/gen.go
```diff
@@ -36,67 +37,89 @@ var genCmd = &cobra.Command{
...
 		if err != nil {
 			log.Logger.Error("experiment build failed")
 			os.Exit(1)
 		}
 
-		// execute template
-		var b bytes.Buffer
-		err = tmpl.Execute(&b, exp)
+		// generate formatted output
+		err = exp.Gen(outputFormat)
```
sriumcp: Reuse Gen

cmd/gen.go
```diff
@@ -36,67 +37,89 @@ var genCmd = &cobra.Command{
...
+	// create text template
+	tmpl, err := template.New("text").Funcs(template.FuncMap{
+		"formatText": formatText,
+	}).Option("missingkey=error").Funcs(sprig.TxtFuncMap()).Parse("{{ formatText . }}")
+	if err != nil {
+		log.Logger.WithStackTrace(err.Error()).Error("unable to parse text template")
+		os.Exit(1)
+	}
+
+	// register text template
```
sriumcp: Example of template registration (replace text with k8s)

cmd/run.go
```diff
@@ -29,14 +26,17 @@ var runCmd = &cobra.Command{
...
+		fio := &FileExpIO{}
+		exp, err := Build(false, fio)
 		log.Logger.Trace("build finished")
 		if err != nil {
 			log.Logger.Error("experiment build failed")
 			os.Exit(1)
 		} else {
 			log.Logger.Info("starting experiment run")
-			err := exp.run()
+			err := exp.Run(fio)
```
sriumcp, line 39: Reuse run. Replace fio with cio for Cluster Ops

cmd/gen.go
```diff
@@ -36,67 +37,89 @@ var genCmd = &cobra.Command{
...
+		// ensure it is a valid template
+		tmpl, err = template.New("tpl").Funcs(template.FuncMap{
+			"toYAML": toYAML,
+		}).Option("missingkey=error").Funcs(sprig.TxtFuncMap()).Parse(string(tplBytes))
+		if err != nil {
+			log.Logger.WithStackTrace(err.Error()).Error("unable to parse template file")
+			return err
+		}
+
+	default:
```
kalantar: I see registration below.  But I don't see the use of the registered items. This should be more like this: 
```bash
templateKey = strings.ToLower(outputFormat)
if templateKey == CustomOutputFormat { // please define reusable constants
    // current custom handling
} else {
    tmpl, ok = builtInTemplates[templateKey]
    if !ok {
        // error handling
    }
}
```

cmd/assert.go
```diff
@@ -64,81 +64,92 @@ var assertCmd = &cobra.Command{
...
+				timeSpent += sleepTime
+			}
+		}
+	}
 }
 
 // completed returns true if the experiment is complete
 // if the result stanza is missing, this function returns false
-func (exp *experiment) completed() bool {
+func (exp *Experiment) completed() bool {
```
kalantar: To implement the `get` function, we need to determine the values of conditions `completed` and `nofailure`. Currently, I am calling `Assert()` twice.  Perhaps, `complete()` and `noFailure()` could be exposed?  Alternatively, there might be a form of `Assert` that returns a list of booleans, one for each condition. We would still want the existing `Assert()` method as well -- it would wrap the new method.

cmd/assert.go
```diff
@@ -64,81 +64,92 @@ var assertCmd = &cobra.Command{
...
+	},
+}
+
+// Assert if experiment satisfies conditions
+func (exp *Experiment) Assert(conditions []string, to time.Duration) (bool, error) {
+	// check assert conditions
+	allGood := true
+	for {
+		for _, cond := range conditions {
+			if strings.ToLower(cond) == completed {
```
kalantar: Could the constants `completed`, `noFailure`. `slos` and `slosByPrefix` be exposed? I create them as well...

kalantar (changes requested):
A few comments.

cmd/gen.go
sriumcp:
```go
tmpl = builtInTemplates["text"]
```

This is where registered items are used. 

Other than reusable constants, the if statement you are suggesting is equivalent to the switch statement above. In any case, since you prefer if, I made it if.

kalantar: I am not concerned about the switch statement. I could only see that "text" was used via an explicit case in the switch. If we dynamically register a template (for example, the k8s template), the switch will not have the key. Instead we need to just try to get it and use or fail.  I wrote it as an if statement since I had just 2 cases, custom and everything else.  Dynamic registration means text is no longer special per se.

kalantar (approved):
lgtm

PR #1056, opened 2021-11-22
V0.8

sriumcp:


kalantar (approved):
I'm approving, but please look at the one comment which questions something you have done. Thank you for simplifying the examples to be command specific.

testdata/iter8.tpl
```diff
@@ -34,7 +34,29 @@ metadata:
   name: {{ .Name }}-{{ $suffix }}
 stringData:
   experiment: |
-{{ . | toYAML | indent 4 }}
+    # task 1: generate HTTP requests for https://example.com
```
kalantar, line 37: Was it the intention to unroll this here. It now differs from the experiment.yaml (cf. run commands 3 and 4 have changed)?

sriumcp: Yes. To be super clear....

`iter8 gen go`  does not require any `experiment.yaml`. It is simply a go  template, and simply executes the template with data  suppplied using `--set`. In other words, you can render any string using this command. This is something we plan to use for code engine (for the time being  at least).

`iter8 gen exp` actually renders an `experiment.yaml` file using an `experiment.tpl` file. Only difference between this and`iter8 gen go` is that it will validate that the output generated through this command is a valid experiment spec. Without this validation, this command is equivalent to the above.

`iter8 manifest` has a built  in template for generating kubernetes manifests. This uses `experiment.yaml` and outputs kubernetes manifests. You are creating this!

`iter8 report` produces reports from experiment results.

***

All of the above have some overlapping template and validation functionality (which have been abstracted into reusable pieces).  But they do very different things from end-user perspective (at the minimum, calling for separate documentation).

PR #1064, opened 2021-12-02
Add community tab

Alan-Cha:


mkdocs/docs/contributing.md
```diff
@@ -59,14 +62,6 @@ that you want to work on it. Something like "I want to work on this" is fine.
 
 ***
 
-## Ask for Help
```
Alan-Cha, line 62: Moved this

mkdocs/mkdocs.yml
```diff
@@ -151,4 +151,5 @@ nav:
     - Iter8 hub: user-guide/topics/iter8hub.md
 - Contributing: contributing.md
 - Roadmap: roadmap.md
-- News/Blog: news.md
```
Alan-Cha, line 154: The slash looks ugly to me. I'd rather revert back to "News".

mkdocs/mkdocs.yml
```diff
@@ -118,7 +118,7 @@ nav:
   - Install Iter8: getting-started/install.md
   - Concepts: getting-started/concepts.md
   - v0.8 tech preview: getting-started/v0.8.md
-  - Community: getting-started/help.md
+  - Get help: getting-started/help.md
```
Alan-Cha, line 121: Typo here.

Additionally, I think the Getting Started tab needs to be reworked. I don't like how there are 5 pages and they all seem kind of disparate. It should be as simple as possible for a new user to get started.

For example, does a new user need to know about the differences between v0.7 and v0.8? I think this may be better suited in the Roadmap tab.

I understand why the install instructions are on its own page (so it can be referenced in other tutorials) but I feel like the Your First Experiment should just repeat the information. It's an additional hurdle for a brand new user to click on a link and go to a new page and come back to the original tutorial.

I also think the there's a lot of overlap between the Get Help page and the Contributing tab. I am also a bit wary of how useful this extra page is as it just lists a few items. It may be cleaner just to reference the Contributing tab section at the end of Your First Experiment.

sriumcp (approved):
lgtm

PR #1066, opened 2021-12-03
CE template with revisions

Alan-Cha:


testdata/ce/go.tpl
```diff
@@ -0,0 +1,67 @@
...
+        - metric: built-in/p95.0
+          upperLimit: 1
+    # task 3: if SLOs are satisfied, do nothing
+    - if: SLOs()
+      run: echo "SLOs are satisfied"
+    # task 4: if SLOs are not satisfied, revert to the previous version
+    - if: not SLOs()
+      run: |
+        # Get the name of the second to last revision
+        penultimateRevisionName=$(kubectl get revisions \
```
Alan-Cha, line 59: This extracts the second to last revision name given the application name via `{{ .CEAppName }}`.

testdata/ce/go.tpl
```diff
@@ -0,0 +1,67 @@
...
+metadata:
+  name: {{ .Name }}-{{ $suffix }}
+stringData:
+  experiment: |
+    # task 1: generate HTTP requests for https://example.com
+    # collect Iter8's built-in latency and error-related metrics
+    - task: gen-load-and-collect-metrics
+      with:
+        versionInfo:
+        - url: https://iter8-ce-demo.fgw1ut94lpp.ca-tor.codeengine.appdomain.cloud
```
sriumcp: Should the URL value be replaced with something like ...

```
{{ .CEAppName }}-{{ .CEAppNamespace }}.ca-tor.codeengine.appdomain.cloud
```

testdata/ce/go.tpl
```diff
@@ -0,0 +1,67 @@
+{{- $suffix := randAlphaNum 5 | lower -}}
+apiVersion: codeengine.cloud.ibm.com/v1beta1
+kind: JobRun
+metadata:
+  name: {{ .Name }}-{{ $suffix }}
```
sriumcp: Could we please get rid of `{{ .Name }}` and replace it with a fixed string, like `experiment-run`.

sriumcp (changes requested):
Please see comments.

The idea of extracting penultimate version inside the experiment is SUPER!

testdata/ce/go.tpl
Alan-Cha: I wonder how useful templating this will be because the URL also contains the location `ca-tor`, right?

sriumcp: What  is this `ca-tor`? Also, does it make sense to template the whole URL field?

sriumcp:
```
- url: {{ .URL }}
```

Alan-Cha: `ca-tor` is the cluster zone. It's Toronto, Canada. But yes, I think it makes more sense to template the entire URL field.

kalantar:
@Alan-Cha @sriumcp is this PR still relevant? It is very old. It predates v0.9.

sriumcp:
Closing since it is not relevant any longer.

PR #1069, opened 2021-12-07
CLI for k8s

kalantar:


kalantar:
Needs test coverage.

Dockerfile.k8s
```diff
@@ -0,0 +1,50 @@
+# Build Iter8
```
sriumcp: I think we need only one Dockerfile. 

Perhaps `mv Dockerfile.k8s Dockerfile`? 

Not sure what the difference is but the move will reveal the changes.

Makefile
```diff
@@ -23,3 +23,14 @@ cmddocs:
 ITER8_BIN ?= /usr/local/bin/iter8
 build:
```
sriumcp, line 24: Shouldn't the build be building `k8s CLI`? Why is it still building `basecli`?

Makefile
```diff
@@ -23,3 +23,14 @@ cmddocs:
...
 	cd clibase && go build -o $(ITER8_BIN)
+
+
+ITER8_IMG ?= iter8/iter8cli:latest
+docker-build:
```
sriumcp, line 29: See above comment on Dockerfile.

Makefile
```diff
@@ -23,3 +23,14 @@ cmddocs:
...
+	docker build -f Dockerfile.k8s -t $(ITER8_IMG) .
+
+docker-push:
+	docker push $(ITER8_IMG)
+
+build-k8s:
```
sriumcp: I believe we want only one build. Whatever is called `build-k8s` should be the default `build`.

If at all we need `make build-basecli`, we will do that eventually, but I don't think this should be there at all (in other words, we should have only one main.go in the root folder which builds the k8s CLI).

cmd/assert.go
```diff
@@ -19,68 +19,79 @@ const (
...
-		}
-		if !allGood {
-			log.Logger.Error("assert conditions failed")
-			os.Exit(1)
-		}
-		return nil
-	},
+var AssertOptions = AssertOptionsType{}
+
+func NewAssertCmd() *cobra.Command {
```
sriumcp, line 37: The `NewAssertCmd` function wrapper does nothing other than add needless indirection.

I think Cobra's default `AssertCmd` var works better than this wrapper which is adding complexity without value.

cmd/docs/main.go
```diff
@@ -42,7 +46,10 @@ hide:
 	cdd := viper.GetString(commandDocsDir)
 
 	// automatically generate markdown documentation for all Iter8 commands
```
sriumcp, line 45: Automatic markdown generation can be removed from basecli and absorbed into k8s cli.

cmd/run.go
```diff
@@ -9,38 +9,43 @@ import (
...
+			return nil
+		},
+	}
+
+	return RunCmd
 }
 
 func init() {
-	RootCmd.AddCommand(RunCmd)
+	RootCmd.AddCommand(NewRunCmd())
```
sriumcp, line 48: See comment above on wrappers.

k8s/cmd/report.go
```diff
@@ -0,0 +1,110 @@
...
+func NewReportCmd(factory cmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
+	o := newReportOptions(streams)
+
+	cmd := basecli.NewReportCmd()
+	var example = `
+# Generate text report for the most recent experiment running in current Kubernetes context
+iter8 report --remote`
+	cmd.Example = fmt.Sprintf("%s%s\n", cmd.Example, example)
+	cmd.RunE = func(c *cobra.Command, args []string) error {
+		if err := o.complete(factory, c, args); err != nil {
```
sriumcp: PreRunE is a better place for `o.complete` and `o.validate`, as opposed to `RunE`, which can be more focused only on the actual run logic. 

Incidentally, what does `o.complete` do?

k8s/cmd/run.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return err
+	}
+
+	return exp.Run(expIO)
+}
+
+func NewRunCmd(factory cmdutil.Factory, streams genericclioptions.IOStreams) *cobra.Command {
+	cmd := basecli.NewRunCmd()
+	cmd.Hidden = true
```
sriumcp, line 17: Nice to see it hidden.

sriumcp (changes requested):
Minor comments.

Dockerfile.k8s
kalantar: only real difference was build target. replaced original.

cmd/assert.go
kalantar: Two instances of the `assert` command (and `run` and `report`) are needed. Both for `iter8 assert` and `iter8 k assert`.  It is not sufficient to reuse the same instance since the options and behaviors are different.  A `NewAssertCmd` method does this. For uniformity, I applied the same pattern everywhere.

k8s/cmd/report.go
kalantar: `complete()` is computing value of variables used by `run()` that either need default values or that can be derived from the input. Most of these are the same; perhaps a common method can be used.

cmd/assert.go
sriumcp: Let us please use the default Cobra way of handling  this using command variables.

Please rename `iter8 k assert` command as `KAssertCmd` or whatever you like, and use it along side `AssertCmd`. 

If you prefer that  I do this in a subsequent PR, I am take this activity on.

We want to use `cobra add`  command functions and be able to use the auto generated command scaffolding created by Cobra  with the least amount of edits. This function wrapper increases the  work needed to reuse Cobra's auto generated scaffolding. Is not really needed and an extra step in the way of creating new commands.

sriumcp: I `git clone`d your repo and did the following.

```
cd iter8/k8s
cobra add test
```

At this point,  I got  a `test.go` file, nicely scaffolded, under `iter8/k8s/cmd`.

It has `var testCmd` initialized to a command inside it.

***

What the above means is that you can simply do `cobra add assert`, `cobra add run`, `cobra add k`, etc.

You will get a runCmd, kCmd, etc. 

You can add `kCmd` to the `basecli.RootCmd` as you  are already  doing.

Your can also add `runCmd` and `assertCmd` to `kCmd`. Automatically. By doing `cobra add run -p kCmd`, and so on ... 

Here is the documentation for cobra's default command generation which simply works for our use-case without the need for doing anything extra.

https://github.com/spf13/cobra/blob/master/cobra/README.md

***

There is absolutely no conflict. In one case, you have `assertCmd` which is a package scoped variable inside the `k8s` package. In another case, you have an `AssertCmd`, which is an exported variable inside the `basecli` package.

***

All this without having to do the extra work of recreating what cobra already creates for you using its builtin `add` functionality.

k8s/cmd/report.go
sriumcp: I saw the complete() code which either is no-op in one case, or replicated in other cases.

So, yes, removing the code replication and making a single function makes sense.

I now understand what complete does. You can keep this in RunE or PreRunE depending upon your preference. Both look equally good places for complete.

cmd/assert.go
kalantar: Let me reiterate that I use this pattern to enable reuse. This eliminates the need to re-implement or duplicate code. I do not redefine all the flags that `cmd/assert.go` defines. Nor do I redefine the name, example or description. Instead, they are defined just once and reused. This is accomplished by using two calls to `NewAssertCmd()` -- one in `cmd/assert.go` and one in `k8s/cmd/assert.go`.

This pattern may not exactly adhere to the "simplest" cobra pattern. That does not mean that it isn't a valid and useful pattern. The differences are minor and they provide significant reuse in a few cases.

A requirement to support `cobra add` as a way to **stub** code is a new requirement. It is not at all clear that the benefits are worth any effort. In particular, any commands that support k8s generic flags need to do a bunch of extra work and will require a rewrite of the stub. That said, naming is irrelevant. Two small changes enabled this feature.

sriumcp (approved):
lgtm

PR #1076, opened 2021-12-14
add retry for secret gets

kalantar:

### DIFF
--- a/cmd/kubernetes.go
+++ b/cmd/kubernetes.go
@@ -86,32 +86,25 @@ func getSecretWithRetry(client *kubernetes.Clientset, ns string, nm string) (s *
 			return s, err
 		}
 		if !k8serrors.IsNotFound(err) {
+			log.Logger.Errorf("unable to read secret: %s; %s\n", nm, err.Error())
 			return nil, err
 		}
 		time.Sleep(GetRetryInterval)
 	}
 	// tried MAX_RETRIES times
-	return nil, fmt.Errorf("experiment \"%s\" not found", nm)
+	log.Logger.Errorf("experiment \"%s\" not fouund; unable to read secret: %s\n", nm, err.Error())
+	return nil, fmt.Errorf("experiment not found")
 }
 
 func GetExperimentSecret(client *kubernetes.Clientset, ns string, id string) (s *corev1.Secret, err error) {
-	// ctx := context.Background()
-
 	// An id is provided; get this experiment, if it exists
 	if len(id) != 0 {
 		nm := SpecSecretPrefix + id
 		s, err = getSecretWithRetry(client, ns, nm)
-		// s, err = client.CoreV1().Secrets(ns).Get(ctx, nm, metav1.GetOptions{})
-		if err != nil {
-			if k8serrors.IsNotFound(err) {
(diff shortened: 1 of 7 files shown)


cmd/kubernetes.go
```diff
@@ -75,13 +79,29 @@ func GetExperimentLogs(client *kubernetes.Clientset, ns string, id string) (err
 	return nil
 }
 
+func getSecretWithRetry(client *kubernetes.Clientset, ns string, nm string) (s *corev1.Secret, err error) {
+	for i := 0; i < MaxGetRetries; i++ {
+		s, err = client.CoreV1().Secrets(ns).Get(context.Background(), nm, metav1.GetOptions{})
+		if err == nil {
```
sriumcp, line 85:
```go
if err != nil {
  could you please log the error here?
}
```

cmd/kubernetes.go
```diff
@@ -75,13 +79,29 @@ func GetExperimentLogs(client *kubernetes.Clientset, ns string, id string) (err
...
-	ctx := context.Background()
+	// ctx := context.Background()
 
 	// An id is provided; get this experiment, if it exists
 	if len(id) != 0 {
 		nm := SpecSecretPrefix + id
-		s, err = client.CoreV1().Secrets(ns).Get(ctx, nm, metav1.GetOptions{})
+		s, err = getSecretWithRetry(client, ns, nm)
+		// s, err = client.CoreV1().Secrets(ns).Get(ctx, nm, metav1.GetOptions{})
 		if err != nil {
```
sriumcp: Could you please log error here?

cmd/kubernetes.go
```diff
@@ -165,7 +185,8 @@ type KubernetesExpIO struct {
 // read experiment spec from secret in the Kubernetes context
 func (f *KubernetesExpIO) ReadSpec() ([]base.TaskSpec, error) {
 
-	s, err := f.Client.CoreV1().Secrets(f.Namespace).Get(context.Background(), f.Name, metav1.GetOptions{})
+	s, err := getSecretWithRetry(f.Client, f.Namespace, f.Name)
```
sriumcp, line 181: Rather than log the error here, it is preferable to log errors the first time they're generated / observed, rather than let them cascade all the way up.
***
This simple pattern ensures that each error is logged exactly once  (not more, and not less).

cmd/kubernetes.go
```diff
@@ -183,7 +204,8 @@ func (f *KubernetesExpIO) ReadSpec() ([]base.TaskSpec, error) {
 // read experiment result from Kubernetes context
 func (f *KubernetesExpIO) ReadResult() (*base.ExperimentResult, error) {
 	resultSecretName := f.Name + "-result"
-	s, err := f.Client.CoreV1().Secrets(f.Namespace).Get(context.Background(), resultSecretName, metav1.GetOptions{})
+	s, err := getSecretWithRetry(f.Client, f.Namespace, resultSecretName)
```
sriumcp, line 198: Could you please log the error inside `getSecretWithRetry` func instead of outside?

sriumcp (changes requested):
Minor changes requested.

sriumcp (approved):
lgtm

PR #1093, opened 2021-12-16
change default image latest to 0.8

kalantar:

### DIFF
--- a/basecli/k8s.tpl
+++ b/basecli/k8s.tpl
@@ -16,9 +16,9 @@
 {{/* -- name ----------------------------------------------- */}}
 {{- $name := printf "experiment-%s" $id -}}
 {{/* -- version -------------------------------------------- */}}
-{{- $version := printf "0.8" -}}
+{{- $version := iter8MajorMinorVersion -}}
 {{/* -- image ---------------------------------------------- */}}
-{{- $image := printf "iter8/iter8:0.8" -}}
+{{- $image := defaultImage -}}
 {{- if hasKey .Values "image" -}}
   {{- $image = .Values.image -}}
 {{- end -}}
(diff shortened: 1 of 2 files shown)


basecli/k8s.tpl
```diff
@@ -18,7 +18,7 @@
 {{/* -- version -------------------------------------------- */}}
 {{- $version := printf "0.8" -}}
 {{/* -- image ---------------------------------------------- */}}
-{{- $image := printf "iter8/iter8:latest" -}}
+{{- $image := printf "iter8/iter8:0.8" -}}
```
sriumcp: Could  we please use the version variable inside Iter8 (used by `iter8 -v` command) instead of hard coding `0.8`?

This is a solution that continues to work for `0.9` as well...

This is useful for both $version and $image.

sriumcp (changes requested):

basecli/k8s.tpl
kalantar: This is hard to do well in a template; we must rely on code to do it. Which makes overriding it (especially for the image; for testing) harder.

sriumcp:
```
{{- $image := defaultImage -}}
{{- if hasKey .Values "image" -}}
  {{- $image = .Values.image -}}
{{- end -}}
```

```
{{- $version := iter8MajorMinorVersion -}}
```

The above is all we need for this to work. You can register the functions `iter8MajorMinorVersion` and `defaultImage` like [this](https://github.com/iter8-tools/iter8/blob/ee99956b0f2feaf0d93ae26e65b3d18f267ddb5f/basecli/report.go#L121).

sriumcp (changes requested):
Please see template/code example for default images and versions.

sriumcp:
This PR partly addresses #1095

sriumcp (approved):
LGTM

PR #1107, opened 2022-01-05
`iter8 gen exp` subcommand

### LINKED ISSUE #1103: Restructure Iter8 v0.8 experiment  folder to be in line with a Helm Chart
Currently, an Iter8 experiment folder does not have any restrictions in terms of its structure (other than the fact that it needs to include an `experiment.yaml` file).

Restructure this so that it resembles a Helm Chart (in other words, Iter8 experiment will be a special type of Helm chart).

***

Key advantages of the above redesign are the following.

1. We can do away with `iter8 gen go` (for which the primary use case is code engine install). Simply modify the templates used for rendering instead to achieve the code engine install.
2. Inherits all the benefits of Helm charts, including the ability to use Helm directly to install experiments, embed experiments as Helm sub charts, and also documentation & versioning for the chart using `Chart.yaml`, and `README.md`, and documentation for templates using `templates/NOTES.txt`.

This does affect a few iter8 subcommands, but mainly `iter8 gen exp` ATM.

sriumcp:
Resolves #1103 

*** 

1. Enables Helm style values (including one or more values file, http values file, and set commands)
2. id & app variables are now shared between basecli and cmd packages (instead of being duplicated)
3. Mkdocs now has the following.
      1. Changes to `your-first-experiment.md` including additional links and improved wording.
      2. Documentation for iter8 hub and `iter8 gen exp` commands. Describes Iter8 experiment charts.
      3. Fixed Roadmap
      4. Removed images that are not needed. Moved them closer to where they are used.
5. Test now includes `iter8 gen exp` command also
      1. Hub location in test needs to change after merge
6. Improved go docs for the `base`, `log`, and `basecli` packages.

***


basecli/exp.go
```diff
@@ -16,64 +16,77 @@ limitations under the License.
...
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/engine"
+	"helm.sh/helm/v3/pkg/getter"
 )
 
 // expCmd represents the exp command
 var expCmd = &cobra.Command{
 	Use:   "exp",
-	Short: "Render experiment template in the file experiment.tpl with values",
+	Short: "Render experiment.yaml file for local experiments",
```
kalantar: why is this only for  *local* experiments?

sriumcp: Fixed.

basecli/k8s.go
```diff
@@ -128,12 +135,33 @@ iter8 gen k8s`,
...
 
+// GetIdFlag returns the id flag.
+// This function enables reuse of this flag across subcommands.
+func GetIdFlag() *pflag.Flag {
+	name := "id"
+	f := pflag.Lookup(name)
+	if f == nil {
+		pflag.StringVarP(&Id, name, "i", "", "if not specified, a randomly generated identifier will be used")
+	}
+	return pflag.Lookup(name)
```
kalantar: Why this pattern? It seems more complex and I am inclined to think we want to observe and error if the parameter already exists. Otherwise, we may miss a conflict.

cmd/assert.go
```diff
@@ -48,8 +48,10 @@ iter8 k assert --id $ID -c completed,nofailure,slosby=0 -t 5s`
 
 		return nil
 	}
-	k8sExperimentOptions.addIdOption(assertCmd.Flags())
-	k8sExperimentOptions.addAppOption(assertCmd.Flags())
+
+	// initialize options for assertCmd
+	assertCmd.Flags().AddFlag(basecli.GetIdFlag())
+	assertCmd.Flags().AddFlag(basecli.GetAppFlag())
```
kalantar, line 54: Why are we adding these to the basecli assert?  Aren't they k8s specific?

basecli/k8s.go
```diff
@@ -39,26 +47,26 @@ func runGetK8sCmd(cmd *cobra.Command, args []string) (err error) {
...
-	// not that if both --app=foo and --set id=bar are used,
-	// the --app optiom will take precedence
-	if len(app) > 0 {
-		values = append(values, "app="+app)
-	}
-
-	err = ParseValues(values, v)
-	if err != nil {
-		return nil, err
+	// note that if both --app=foo and --set id=bar are used,
```
kalantar: documentation should be  `.... --set app=bar are used.`

basecli/k8s.go
sriumcp: Order in which init() functions are called in go is non-deterministic. 

So,  there is no real way  to  make sure the flag is added in `k8s.go` before it can be retrieved in `iter8 k assert.go` or `delete.go` for instance.

This pattern makes the order of init invocations irrelevant. If the flag is not already part of the default `pflag` flagset, it will simply be added and then returned, which is the desired behavior.

basecli/k8s.go
sriumcp: Fixed.

mkdocs/docs/getting-started/your-first-experiment.md
```diff
@@ -4,57 +4,65 @@ template: main.html
...
-[Iter8 experiments](concepts.md#what-is-an-iter8-experiment) are specified using the `experiment.yaml` file. The `iter8 run` command reads this file, runs the specified experiment, and writes the results of the experiment into the `result.yaml` file.
-
-Run the experiment you downloaded above as follows.
-
+## 3. Generate `experiment.yaml`
+Generate the `experiment.yaml` file which specifies your load test experiment.
 ```shell
 cd load-test
-iter8 run
+iter8 gen exp --set url=https://example.com
```
kalantar, line 29: We can still start with an `experiment.yaml`. Is using `iter8 gen exp` as part of the first experiment maybe more complex than it needs to be? I agree that the chart is flexible. But it seems to suggest that `iter8 gen exp` is always needed. Is this one more concept that we don't need to start?

kalantar: If *every* experiment is a helm chart, then perhaps there is no need for the `gen` command; just make `run` do the generation.

basecli/k8s.go
```diff
@@ -56,14 +56,14 @@ func Generate() (result *bytes.Buffer, err error) {
...
+	// note that if --app=foo and --id=bar are used,
 	// the --id option will take precedence
 	if len(Id) > 0 {
 		v["id"] = Id
 	}
 
 	// add app=app if --app option is used
-	// note that if both --app=foo and --set id=bar are used,
+	// note that if --app=foo and --id=bar are used,
 	// the --app option will take precedence
```
kalantar: These comments still do not make sense to me. I think we are saying:

- if `--id=foo` and `--set id=bar`; then `--id` option takes precedence
- if `--app=foo` and `--set app=foo`; then `--app` option takes precedence

basecli/k8s.go
kalantar: I do not understand how the flag being added in `k8s.go` and in `delete.go` are related to each other. These are different cobra `Command`s. Qhy is the order of `init()` between them relevant?

mkdocs/docs/getting-started/your-first-experiment.md
sriumcp: I would like the experiment chart to be flexible enough so that most users don't ever need to understand or look into `experiment.yaml` or tasks or any such thing... but simply reuse with different values.
***
Definitely more documentation is needed for supported values, which I plan to include before the blog writeup.
***
So, yes, the concept we want to push is the templated experiment, not the experiment.yaml.

sriumcp: Yes, we can make `run` do the generation directly (we will explore this option shortly).
***
At the moment, `gen k8s` seems to require `experiment.yaml`. It doesn't seem right to `run` locally first just for the sake of generating the `experiment.yaml`.

basecli/k8s.go
sriumcp: Sorry. Resolving again after fixing.

mkdocs/docs/getting-started/your-first-experiment.md
sriumcp: We can eventually consider your suggestion captured in this issue now: https://github.com/iter8-tools/iter8/issues/1108

basecli/k8s.go
sriumcp: Both `iter8 gen k8s` and `iter8 k delete/assert/...` make use of `id` and `app` variables and flags.

The pattern above ensures that:
1. App and Id variables are defined in one place and reused across these subcommands.
2. Flags are available for these subcommands to set these variables through command line.
3. The flags themselves are reused. In other words, there is a single pflag.Flag object, included as part of the pflag.Flagset, and looked up by whoever needs it. Reuse of flags ensures that they don't end up with different descriptions in different places (for example).
4. Both the initial addition to the flagset and the lookups happen as part of the init functions. But there is no way to control the order of init function executions. This means, you cannot simply place the addition of the flag to the flagset in one init func, and expect to see it for sure in another init func. The pattern used above ensures that no matter which init func is called first, the flag will be added to the flagset, once and exactly once,  and then will be reused subsequently.

sriumcp: Similar to the idea of using NewCmd() patterns for reusing command blocks, this also reuses the flags passed to those command blocks.

sriumcp: If you do have a simpler pattern/alternative to achieve 1, 2, 3 and 4 above -- we can certainly consider it!

sriumcp: The following alternative should work as well... but its a bit of careless coding, though simpler! This means, the app flag is recreated in the pflag flagset, every time any command "gets" it, which seems... wasteful. I don't really think there is a big diff either way.

```go
 // GetAppFlag returns the app flag.
 // This function enables reuse of this flag across subcommands.
 func GetAppFlag() *pflag.Flag {
 	pflag.StringVarP(&App, "app", "a", "", "label to be associated with an experiment, default is 'default'")
 	return pflag.Lookup("app")
 }

sriumcp (commented):
added further comments.

basecli/k8s.go
sriumcp: @kalantar I now understand that the issue is a bit more deeper than docs and code structure ... 

1. `iter8 gen k8s` current does not (rightly so) ignore either id or app when they are specified. However `iter8 k *` commands current do ignore app when id is specified. So far so good.
2. This means the app flags are actually different across these two command groups (because the behavior of the app flag is different).
3. This means, we don't need `GetAppFlag`. `GetIdFlag` still makes sense, and the way App and Id variables are reused still makes sense.

I will push these changes shortly.

basecli/k8s.go
sriumcp: Nothing is ignored in `iter8 gen k8s`. Will push a change to reflect this shortly. (please see earlier conversation).

sriumcp: Fixed now.

basecli/k8s.go
sriumcp: Fixed now.

mkdocs/docs/getting-started/your-first-experiment.md
kalantar: > At the moment, gen k8s seems to require experiment.yaml. It doesn't seem right to run locally first just for the sake of generating the experiment.yaml.

Agreed. `gen k8s` would have to change as well, I think.

kalantar (approved):
lgtm

PR #1111, opened 2022-01-07
Release binaries with appropriate version information

sriumcp:
This PR also dispenses with `iter8 -v` and instead adds the `iter8 version` subcommand so that versions can be tracked properly in the binaries.


kalantar (approved):
looks good

PR #1123, opened 2022-01-11
Hiding `iter8 gen exp` command

### LINKED ISSUE #1108: Consider getting rid of `iter8 gen exp`
Every experiment is intended to be run through an experiment chart. Consider rolling `iter8 gen exp` into `iter8 run`.

In particular, `iter8 run` should:
1. Generate `experiment.yaml` from current chart and run it. For this purpose, it would take exactly the same set of options as `iter8 gen exp`, including ways to set values. This combines the behaviors of current `iter8 gen exp` and `iter8 run`.
2. Support `iter8 run --dry` flag to simply generate `experiment.yaml` without running the experiment (i.e., this performs what the current `iter8 gen exp` does).

sriumcp:
Resolves #1108


kalantar (approved):
Looks good; we will need an issue for `iter8 gen k8s`, won't we?

sriumcp:
> we will need an issue for iter8 gen k8s, won't we?

Yes, absolutely. I think we need to do a bit more for launching experiments into Kubernetes (in particular, what happens when you launch experiments repeated for the same app)... we need an issue/design discussion. Will open these shortly.

Issue #1122, opened 2022-01-10
Create tutorials that demonstrate various facets of load testing


sriumcp:
Iter8's load testing template already supports the following three aspects of testing.
***
- [x] Fine grained control over the request generation
- [x] HTTP GET & POST endpoints
- [x] Fine grained control over how errors are defined, and which percentiles are computed  and used in SLOs
***
Create tutorials that clearly document how the user is expected to exercise these features. Use `--set` values or `values.yaml` for tutorials. The user should not have to understand experiment.yaml` or tasks.
***
Also create:
- [x] A started tutorial for Knative intended to be submitted linked from the Knative repo

PR #1133, opened 2022-01-13
Improved tutorials for load testing

### LINKED ISSUE #1131: Write result of `iter8 run --dry` to stdout
`iter8 run --dry` renders experiment.yaml without running the experiment. Use of `iter8 run --dry` is to see the experiment being generated; it may be easier for users if we (also) output to stdout. This would be behavior similar to other tools like `helm template`.

### LINKED ISSUE #1132: iter8 run --dry needs to be more verbose about problems
After iter8 hub -e load-test, if I don't give a valid url parameter in values.yaml, it seems iter8 run --dry would complete without generating an experiment.yaml file and without any error. This seems a bit oddd

sriumcp:
1. Fixes #1132 
2. Fixes #1131 
3. Addresses most of #1122 


kalantar (approved):
lgtm

Issue #1146, opened 2022-01-19
Create a GitHub Actions task


sriumcp:
1. Needs to show up in Marketplace (hosted in the main Iter8 repo for star-ability; repo structure)
2. Figure out semantic versioning of the action, and also implications for version breaking Iter8 changes (the actions version needs to subsume this story)
3. Identical in functionality to the Tekton -- except it works inside GH actions pipelines

kalantar:
Current POC is here: https://github.com/kalantar/iter8-run

kalantar:
I've updated the POC to use the latest version of iter8. In particular, the set of configurable properties is larger and `--set` is used instead of attempting to manipulate `values.yaml`.

Further, I moved the definition (`action.yaml`) into a subdirectory `load-test` to verify that it did not need to be at the root.

kalantar:
GitHub seems to recognize that a project is defining an action. Similar to the way it prompts you to create a PR from a recent branch, it prompts "Publish this Action to Marketplace"

kalantar:
https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace

suggests that the action.yaml should be in the root repository and only one defined. However, there are others with multiple and non root. We should try for this; I think there will be different actions for different experiment types.

kalantar:
@sriumcp is there a way to specify `errorRanges` and `percentiles` via `--set`?

sriumcp:
I believe there is, but I need to study `helm` a lot more to describe this correctly.

These are exactly the fields for which I wouldn't use `--set` and would use a yaml file instead and pass that using `-f`.

sriumcp:
Also, I think there is no real need for an end-user to tinker with them (and in the rare occasion that they do, values.yaml is an acceptable solution).

kalantar:
> I believe there is, but I need to study `helm` a lot more to describe this correctly.
> 
> These are exactly the fields for which I wouldn't use `--set` and would use a yaml file instead and pass that using `-f`.

There is no real need to investigate further. If this we do not expect to use `--set` then we do not need a GH action input for it.  Instead we can introduce an input for the values file.  It should be there anyway.

kalantar:
Any opinions?  Individual input variables or a comma separated list:

```yaml
uses: action
with:
   error_rate: 0
   mean_latency: 100
   p90: 150
```

vs:

```yaml
uses: action
with:
  slos: error-rate=0,mean-latency=100,p90=150
```

[...]

PR #1150, opened 2022-01-21
load-test action

kalantar:
Adds GH action iter8-tools/iter8/load-test.  See https://github.com/iter8-tools/iter8/issues/1146.


kalantar:
Can be tried using action `iter8-tools/iter8/load-test@pull/1150/head` (ie, reference the PR).

kalantar:
It is not clear we should use a sub-directory as we do here.  https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace suggests 1 action per repository and in the root of the project. On the other hand, it does work to use a sub-directory and there can be multiple actions per repo (see, for example, https://github.com/github/codeql-action).

If we have multiple actions (for example, for different experiment types) we might want an "actions" sub-directory as a way to organize them. However, this would have the negative effect of making references longer.

I suspect that the `README` should be combined with that at the root level.

sriumcp:
This is starting to look very neat. 
***
However, I think we have Iter8 wide semantic versioning issues which we will want to sort out before merging this action. I am not sure what will be the impact on this action (hopefully none), but it is better to at least understand the semver implications before this merge.
***
Meanwhile, a couple of questions...

1. How are you planning to test this action? What is the recommended approach for testing GitHub actions? How do others (good packages) test their GitHub actions? I didn't see any tests included, but assume they are WIP.
2. What is the issue with "." vs "_"? Helm `--set` also has issues with dotted values. But the workaround there seems to be `'90\.5'`, i.e., use a quoted string with an escaped dot. Is there anything cheap  with GH actions as well instead of the long if/else switches?

load-test/entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+  OPTIONS="$OPTIONS -f ${INPUT_VALUES}"
+fi
+
+echo "Create experiment.yaml for inspection"
+echo "$ITER8 run --dry $OPTIONS"
+$ITER8 run --dry $OPTIONS
+cat experiment.yaml
+
+echo "Run Experiment"
+LOG_LEVEL=trace $ITER8 run $OPTIONS
```
sriumcp: LOG_LEVEL needs to be  an input to the action?

load-test/entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+  exit 1
+fi
+
+# no need to cleanup
+
+echo "Fetch load-test experiment"
+$ITER8 hub -e load-test
+cd load-test
+
+OPTIONS=""
```
sriumcp: Is it possible to have a map[string]string as input in GH actions? And then simply iterate  over the keys and values of the map and set  them all? Instead of trying to statically check for values to be set?

sriumcp (commented):
Initial comments.

load-test/README.md
```diff
@@ -0,0 +1,71 @@
...
+    error_rate: 0
+    mean_latency: 50
+    p90: 100
+    p97_5: 200
+```
+
+Set the number of requests sent during the load-test to 200, the number of requests per second to 10, and the number of parallel connections used to send the requests to 5:
+
+``` yaml
+- uses: iter8-tools/load-test@v1
```
sriumcp: How/where do you set @v1 in the action? 

Also, what is the way to set it as `v0.8` instead of `v1`? I have seen many GH actions using `v1`, `v2`, ... perhaps, this is not intended to be a semver tag?

sriumcp (commented):
More comments.

load-test/README.md
sriumcp: Also... if there's a breaking change to Iter8 CLI (which will cascade into the Iter8 image also).

What happens to pipelines which currently use this action?
***

Less obvious ... if there's a breaking change to the experiment chart...

What happens to pipelines which currently use this action?
***
The above two questions are part of the Iter8-wide semver choices we need to make.

load-test/entrypoint.sh
sriumcp: I read the input syntax. Looks like this is not possible.

sriumcp: How about:

```yaml
- uses: iter8-tools/load-test@v1
  with:
    valuesFile: path/to/my-values.yaml
```

For CI/CD use-cases, it makes total sense for the user to check in the values file as part of the repo.

Of course, this would also enable users to write the following (creating file on the fly; also, completely eliminating the need to escape special character .):
```yaml
- run: |
    cat << EOF > path/to/my-values.yaml
    url: https://example.com
    payloadURL: https://payload.com
    SLOs:
      p95.0: 200
      error-rate: 0
    EOF
- uses: iter8-tools/load-test@v1
  with:
    valuesFile: path/to/my-values.yaml
```

kalantar:
> How are you planning to test this action? What is the recommended approach for testing GitHub actions? How do others (good packages) test their GitHub actions? I didn't see any tests included, but assume they are WIP.

I have not attempted this yet beyond testing it manually with another project. I will investigate best practice and agree we need. For now, this can help stimulate further design considerations.

kalantar:
> What is the issue with "." vs "_"? Helm --set also has issues with dotted values. But the workaround there seems to be '90\.5', i.e., use a quoted string with an escaped dot. Is there anything cheap with GH actions as well instead of the long if/else switches?

The inputs get translated to environment variable names Environment variable names cannnot have `.` in them. Hence the use of `_`.  The values can also be provided to the command as arguments. I do not know if this approach would make the name irrelevant. Will explore this.

Did not see anything that would make the if/else any better.

load-test/entrypoint.sh
kalantar: done

load-test/README.md
kalantar: > How/where do you set @v1 in the action?
> 
> Also, what is the way to set it as `v0.8` instead of `v1`? I have seen many GH actions using `v1`, `v2`, ... perhaps, this is not intended to be a semver tag?

The value placed after the `@` is a github reference. If you use `@main` it uses the main branch.  If you use `@pull/1150/head` it will use this pull request.  If you use `@v0.8.x` it will use that release.

For clarity (and desire not to change frequently), I think it desirable to document it as `@v0.8`. However, I am not sure that this works from a versioning perspective.  If we use `@v0.8.x` I think any issues of incompatibility go away.

kalantar: Creating the action in a separate project would allow for semantic versioning that does not exactly match the version of iter8; ie, `v0.8`. However, this introduces (a) a new repo and (b) the semantic versioning questions.

load-test/entrypoint.sh
kalantar: > For CI/CD use-cases, it makes total sense for the user to check in the values file as part of the repo.
> 
> Of course, this would also enable users to write the following (creating file on the fly; also, completely eliminating the need to escape special character .):

I completely missed this on first reading. I agree. And the current implementation takes an input `values` that can be used in just this way. Question is ... **do we want to completely eliminate the other options**?

Doing so would simplify the implementation (no need to worry about these inputs) and also simplifies the documentation challenge. Currently, we require descriptions in `action.yaml` and documentation in `README.md` both of which must be consistent with the helm chart.

sriumcp: I think `helm` style `--set` values is starting to make more and more sense for the command line, and less and less sense in this GitHub action, where `values.yaml` looks like the easier option all around (both for the end-user and for the maintainers).

kalantar: made these changes.

kalantar:
Updated to latest version of chart and added working test case.

kalantar:
Updated for v0.9

action.yaml
```diff
@@ -8,6 +8,10 @@ inputs:
   valuesFile:
     description: 'Path to file of configuration values.'
     required: false
+  validateSLOs:
```
sriumcp, line 11: Looking good.

kalantar:
supports chartVersion constraint
added "branding": green check-circle; cf. https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#branding and https://haya14busa.github.io/github-action-brandings/

kalantar:
@sriumcp this is ready for review

action.yaml
```diff
@@ -0,0 +1,28 @@
+name: 'Run Iter8 Experiment'
+description: 'Run an Iter8 experiment'
+author: Iter8
```
sriumcp, line 3: I think its perfectly ok to say "Michael Kalantar" here  (along with the email you use to sign off). Its up to you. Iter8 authors is also perfectly ok.

README.md
```diff
@@ -18,21 +18,52 @@
...
+
+``` yaml
+    - run: |
+        cat << EOF > expConfig.yaml
+          url: https://httpbin.org/get
+          SLOs:
+            http/error-rate: 0
+            http/mean-latency: 100
+        EOF
+    - uses: iter8-tools/iter8@v0.9
```
sriumcp, line 33: Is this simply a docker file ref? Or a git ref?

Could you help understand the components of this token `iter8-tools/iter8@v0.9`? How is `v0.9` used to get the "right version" of Iter8?

What are the steps involved when we release v0.10.0 of Iter8  CLI and want to update the GH action also to v0.10?


***

README.md
```diff
@@ -18,21 +18,52 @@
 
 ## :rocket: Getting started
 
-### Install Iter8 CLI
+### Using Iter8 Action
```
sriumcp, line 21: Could you please have two sections under Getting started?

### Local  usage (where we want to retain the previous documentation like Install Iter8, HTTP and gRPC load tests), and 

### GitHub Actions usage (where we want to add your new documentation).

Eventually, we will also include Kuberenetes usage above.

sriumcp (changes requested):
Please see one request for change, one question and a suggestion.

README.md
sriumcp, line 33: There doesn't seem to be any connection between `iter8-tools/iter8@v0.9` used here and `iter8/iter8:0.9` used in the Docker file.

***

This is what we want to connect up, so that when we go to v0.10.0 of Iter8 CLI, the GH action can be upgraded easily/automatically.

sriumcp (commented):
More comments.

README.md
kalantar: It is meant here as a github ref. The formati is `{owner}/{repo}@{ref}` It can be many things. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses
I expect that if e release v0.10.0 we would the action reference to v0.10. We would also update action/Dockerfile to use iter8:v0.10

README.md
kalantar: I guess I think I do have these sections albeit with different titles and in the reverse order. The order should reflect the common use for people who read the README. Of course, since I wrote actions, I think it the more common use case :) Practically speaking, github action users will come via the readme since that is what the marketplace will give them. Will CLI users start with readme or doc?

README.md
kalantar: By using action v0.9 one is using that github ref to fetch the dockerfile. So they are connected. Changing the ref in the dockerfile means one will need to change the action ref too. i would prefer less of a relationship so less places to change when changing version but I don't see it.

sriumcp: > By using action v0.9 one is using that github ref to fetch the dockerfile.

How? The Dockerfile ref is hardcoded and doesn't seem dependent on the `v0.9` tag -- which at the moment is not even a valid GitHub ref for the Iter8 repo.

sriumcp: > i would prefer less of a relationship so less places to change when changing version but I don't see it.

Agreed. But you need a valid GitHub ref to accomplish the above. `v0.9` is not that ref ATM, because Iter8 doesn't have this ref.

sriumcp: Perhaps the simplest solution is to say:

```
- uses: iter8-tools/iter8 # you can also replace this with iter8-tools/iter8@tag where tag is any valid Iter8 release tag
```

sriumcp: List of valid Iter8 tags:  https://github.com/iter8-tools/iter8/tags
List of valid Iter8 branches: https://github.com/iter8-tools/iter8/branches

README.md
sriumcp: The CLI usage is IMO, the more predominant usage. We can always reassess thee order. If GH actions does turn out to be more predominant, we can switch. But we should start with the simplest, which is CLI.

README.md
sriumcp: This will not work.
***
If we release `v0.10.0`, then `v0.10` does not automatically become a valid git ref. Only `v0.10.0` is a valid git ref.
***
This means, `iter8-tools/iter8@v0.10` is an invalid action, but `iter8-tools/iter8@v0.10.0` is the valid action.

sriumcp: I  saw https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses.

It says, major version is ok. But it doesn't say minor version is ok. It does say specific version is ok.

README.md
sriumcp: The README is not really meant exclusively for GitHub Actions users. The README will also determine... star worthiness or fork worthiness or use worthiness of the project as a whole.

Since we decided to club the GH Action as part of this repo, it is not right to look at the README in isolation only w.r.t. GH Action.

***

Anyway, please use the section titles ## Local Usage and ## GitHub Actions Usage, which would make it convenient for us to add ## Kuberntes Usage (the last is coming soon).

kalantar: Then local first.

README.md
kalantar: > But you need a valid GitHub ref to accomplish the above. v0.9 is not that ref ATM, because Iter8 doesn't have this ref.

Ah. My mistake. This is an issue. I suppose we can create a branch and keep it up to date on release. But this is a significant complexity.

kalantar:
After discussion, decision to create https://github.com/iter8-tools/iter8-action

PR #1151, opened 2022-01-26
gRPC load test and other enhancements

sriumcp:
1. Bumping up code coverage to 69%.
2. Removing tag based validation `validate:"url"` -- this is turning out to be really complex beyond the most elementary type of validation
3. Removing insight types -- this was an overkill  and not required
4. Removing task specs (Tasks can be directly serialized and deserialized now)
5. GRPC metrics collection task (new task)
6. renaming the old collect task as collect_http, along with cleanup
7. Introducing the notion of a sample metric type, and aggregation operations on top of it
8. Removing interpolation from run task (this is an excellent example of needless user-facing complexity)
9. Improved reporting using text and html tpl files that are now better aligned, and also using text and html templates (latter is supposed to be more secure while generating html output)
10. Renaming to `gen-load-and-collect-metrics-http`


Alan-Cha (dismissed):
Looks good to me

PR #1152, opened 2022-01-27
Grpc load test chart

sriumcp:
1. Dropping the "preview" tag from 0.8. We now have HTTP and gRPC and a sane semantic versioning mechanism (at least for local experiments)
2. Improved (hopefully, more star-worthy) README.md
3. Fixes to collect_grpc.go task, load-test-grpc chart, a tutorial that demonstrates load testing of a unary grpc service, unit and e2e tests for this feature
4. Retaining qps for http and rps for grpc (since that is the naming conventions followed by fortio and ghz libraries respectively)
5. Metrics utilities (e.g., GetMetricsInfo for both aggregated and non-aggregated metrics)
6. Exposing uniq (for slices) -- useful in many places
7. Better reporting for metrics (both scalar and aggregated metrics will show up)
8. Creating `load-test-http` (essentially, a replica of `load-test`, which is now deprecated). Support for `load-test` should be removed in `v0.10`. `load-test` is ambiguous, given both `load-test-http` and `load-test-grpc`.
9. Cleaned up concepts
10. Dedup `Contributing.md`. It appears once in the repo top level and also in mkdocs 
11. User guide focuses on end-user commands. Other end user concepts to be added as needed. Developer topics to be documented in the wiki (as we're doing already), with links in the `contributing` page


CONTRIBUTING.md
```diff
@@ -12,4 +12,158 @@ our project that we need help with including:
...
+We welcome many different types of contributions including:
+
+* Tutorials
+* Other documentation
+* Builds, CI
+* Tasks from which experiments are composed
+* Bug fixes
+* Web design
+* Communication, social media, blog posts
+* Reviewing pull requests
```
kalantar, line 28: Is it worth explicitly mentioning "bug reports"?

README.md
```diff
@@ -11,9 +7,44 @@
 [![Slack channel](https://img.shields.io/badge/Slack-Join-purple)](https://join.slack.com/t/iter8-tools/shared_invite/zt-awl2se8i-L0pZCpuHntpPejxzLicbmw)
 [![Community meetups](https://img.shields.io/badge/meet-Iter8%20community%20meetups-brightgreen)](https://iter8.tools/0.7/getting-started/help/#iter8-community-meetings)
```
kalantar: Update meetup link to be `0.8` ?

hub/load-test/Chart.yaml
```diff
@@ -19,4 +19,5 @@ maintainers:
   url: https://researcher.watson.ibm.com/researcher/view.php?person=us-spartha
 icon: https://github.com/iter8-tools/iter8/raw/master/mkdocs/docs/images/favicon.png
 appVersion: "0.8"
+deprecated: true
```
kalantar, line 23: why don't we just remove `load-test` entirely? It was a preview.

mkdocs/docs/roadmap.md
```diff
@@ -7,17 +7,11 @@ hide:
...
-
-4. Iter8 experiments & progressive delivery for multi-cluster and edge cloud
-5. Ability to run Iter8 experiments as a step in CI/CD/GitOps pipelines
-6. Auto triggering experiments
-7. `read-external-metrics` task
-8. `recommend-traffic-split` task
-9. Ability to run multiple loops of an experiment
+1. Examples for experimenting inside K8s
+2. A/B(/n) testing example
+3. Iter8 experiments & progressive delivery for multi-cluster and edge cloud
```
kalantar: is there an order? I imagine it would be our current thinking of priority. This changes daily of course, but for now perhaps item 3 should go down the list some.

mkdocs/docs/tutorials/load-test-grpc/unary.md
```diff
@@ -0,0 +1,117 @@
...
+    
+***
+
+???+ note "Before you begin"
+    1. [Install Iter8](../../getting-started/install.md).
+    2. Complete the [quick start tutorial](../../getting-started/your-first-experiment.md).
+
+## 1. Run sample app
+Choose a language, and follow the linked instructions to run the gRPC sample app. 
+
```
kalantar, line 18: Is it necessary to give instructions for so many alternatives? They are all remote references anyway. Perhaps just a reference to the list of available options and do just one?

mkdocs/overrides/main.html
```diff
@@ -32,10 +32,6 @@
 
 {% endblock %}
 
-<!-- Announcement bar -->
-{% block announce %}
-Welcome to the preview release of <strong>Iter8 v0.8</strong>. This release greatly simplifies release engineering experiments for Kubernetes apps and ML models.
```
kalantar, line 37: If we are really removing the preview tag, this announcement should be changed.

kalantar (commented):
> I believe it would be helpful to provide an example of how the user should express a valid value for ITER8HUB.

Please see `iter8 hub -h` and the auto-generated md file for it.

> I wonder about the naming of metrics. Is the `built-in` needed?  Could we make it a default?  Or eliminate it? We have this prefix to identify the provider. However, the provider is really identified by the collection task(s). Do we need to make it explicit in the assessment task as well? If we can eliminate it perhaps `grpc-error-rate` and `http-error-rate` could be `grpc/error-rate` and `http/error-rate`; that is, allow slashes in name.

Could we please revisit this in 0.9 (or later)? Perhaps after we create the read-metrics-from-db task, and get some experience with it in the community? I am open to revisiting the name... but this touches a lot of different pieces (chart templates, tasks, docs, ...) ... significant effort ... and it would be nice to know we're investing in this for the right reasons.

mkdocs/overrides/main.html
sriumcp: Done.

mkdocs/docs/tutorials/load-test-grpc/unary.md
sriumcp: The multiple options make it easy for developers across different languages to get started quicker (so, really good from end-user adoption point of view). 

Besides, I am referring to upstream documentation, and not creating this documentation myself (so, really good from my -- maintainer -- point of view).

Given how simple it is for users from different languages to get started, I think this is really worth documenting.

mkdocs/docs/roadmap.md
sriumcp: Done.

hub/load-test/Chart.yaml
sriumcp: We've already submitted a Knative PR (which got merged) with the load-test chart. So, simply dropping it (preview or otherwise), doesn't feel right. What happens to users who, lets say, already built a CI/CD pipeline with the load-test chart? 
***
The above concern is a bit theoretical at this point... and I have already changed the link/docs to point to `load-test-http`... However, I think we want to get the deprecation process right. The `load-test` deprecation gives us an opportunity to do so.

README.md
sriumcp: Done.

CONTRIBUTING.md
sriumcp: This was really a template I took from CNCF... perhaps, leave this as it is for now? Note that nothing has changed in terms of our contributing docs. Previously, we had a md file in mkdocs, and another in the GitHub repo. This change merely de-dupes them, and ensures it is maintained only in the GitHub repo (and automatically pulled into mkdocs).

kalantar (approved):

PR #1184, opened 2022-03-03
Add collect database task

Alan-Cha:


base/collect_database.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+type Params struct {
+	Name  string `json:"name" yaml:"name"`
+	Value string `json:"value" yaml:"value"`
+}
+
+// collectDatabaseInputs holds all the inputs for this task
+//
+// Inputs for the template:
```
Alan-Cha, line 53: I'm not really sure where I should document all of these inputs so I will put it here for now.

Alan-Cha:
I see the unit test has failed but it's not clear to me what went wrong. I'll take a closer look tomorrow.

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+type Metric struct {
+	Name         string   `json:"name" yaml:"name"`
+	Description  string   `json:"description" yaml:"description"`
+	Type         string   `json:"type" yaml:"type"` // TODO: make enum
+	Units        string   `json:"units" yaml:"units"`
+	Params       []Params `json:"params" yaml:"params"`
+	JqExpression string   `json:"jqExpression" yaml:"jqExpression"`
+}
```
kalantar: which of these fields are required vs optional. Currently they all appear to be optional. I would expect description and units to be optional (though units might have a default) and that params could be omitted when empty.

base/collect_database.go
kalantar: It seems like these are ce specific. Shouldn't it be documented in the ce template?

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+	for _, metricFilePath := range metricFilePaths {
+		for _, versionInfo := range t.With.VersionInfo {
+			// add ElapsedTime
+			elapsedTime, err := getElapsedTime(versionInfo, exp)
+			if err != nil {
+				return err
+			}
+			versionInfo["ElapsedTime"] = elapsedTime
+
+			fmt.Println(elapsedTime)
```
kalantar: Is this for debugging? Use logging instead.

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// collect metrics for all metric files and versionInfos
+	for _, metricFilePath := range metricFilePaths {
+		for _, versionInfo := range t.With.VersionInfo {
+			// add ElapsedTime
+			elapsedTime, err := getElapsedTime(versionInfo, exp)
+			if err != nil {
+				return err
+			}
+			versionInfo["ElapsedTime"] = elapsedTime
```
kalantar: Please use constant for strings like this. Avoids spelling mistakes and easier to modify.

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+			}
+
+			for _, metric := range metrics.Metrics {
+				// perform database query and extract metric value
+				value, ok := queryDatabaseAndGetValue(metrics, metric)
+
+				// check if there were any issues querying database and extracting value
+				if !ok {
+					continue
+				}
```
kalantar, line 276: should we be logging an error here or is this a normal occurrence?

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+
+				// convert value to float
+				valueString := fmt.Sprint(value)
+				floatValue, err := strconv.ParseFloat(valueString, 64)
+				if err != nil {
+					log.Logger.Error("could not parse string \""+valueString+"\" to float:", err)
+					continue
+				}
+
+				exp.Result.Insights.updateMetric(metricName, mm, 0, floatValue)
```
kalantar: Shouldn't this be `err = `exp.Result....?

base/collect_database.go
```diff
@@ -0,0 +1,311 @@
...
+func getElapsedTime(versionInfo map[string]interface{}, exp *Experiment) (int64, error) {
+	// ElapsedTime should not be provided by the user
+	if versionInfo["ElapsedTime"] != nil {
+		return 0, errors.New("ElapsedTime should not be provided by the user in VersionInfo: " + fmt.Sprintf("%v", versionInfo))
+	}
+
+	// set StartingTime based on VersionInfo or start of the experiment
+	var startingTime int64
+	var err error
+	if versionInfo["StartingTime"] != nil {
```
kalantar: Please use const for strings used for indexing things.

kalantar:
@Alan-Cha Do we have some examples of how this task would work in an experiment?  Do we have any example experiments. I looked at the test cases but did not understand how the versionInfo is being defined.

base/collect_database.go
Alan-Cha: Okay, that makes sense. I'll look into this.

base/collect_database.go
Alan-Cha: Oops, sorry!

base/collect_database.go
Alan-Cha: Also makes sense.

base/collect_database.go
Alan-Cha: `queryDatabaseAndGetValue()` logs any errors. The mean reason for this check is to avoid logging a second redundant error on line 266.

base/collect_database.go
Alan-Cha: Good catch.

base/collect_database.go
Alan-Cha: Yes, it is CE specific so it probably should not be here. It is documented in the CE template as well but I feel its usage is not clear enough and additional documentation is needed.

base/collect_database.go
Alan-Cha: I'm not entirely sure how to mark which fields are required/optional in code. I see examples using the `omitempty` struct tag in other parts of our code base but I'm not sure if that is what you are asking for. Can you give me some clarification?

kalantar: To mark a field optional, one includes the `omitempty` tag. This indicates that when this field is not set, it should omitted from any yaml/json. However, it is necessary to support the lack of a value. So the go data type for optional fields should be pointers. And their use should verify that they are not `nil`.

base/collect_database.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/itchyny/gojq"
+	log "github.com/iter8-tools/iter8/base/log"
+
+	"text/template"
+	"time"
+
+	"sigs.k8s.io/yaml"
+)
+
+type CollectDatabaseTemplateInput struct {
```
sriumcp: This type doesn't appear to be used anywhere. Should be removed?

base/collect_database.go
```diff
@@ -0,0 +1,315 @@
...
+
+type CollectDatabaseTemplate struct {
+	Url      string            `json:"url" yaml:"url"`
+	Headers  map[string]string `json:"headers" yaml:"headers"` // TODO: should this be map[string]interface{} instead?
+	Provider string            `json:"provider" yaml:"provider"`
+	Method   string            `json:"method" yaml:"method"`
+	Metrics  []Metric          `json:"metrics" yaml:"metrics"`
+}
+
+type Metric struct {
```
sriumcp, line 33: Shouldn't there also be a `Body string` field for POST queries?

base/collect_database.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// collect metrics for all metric files and versionInfos
+	for _, metricFilePath := range metricFilePaths {
+		for _, versionInfo := range t.With.VersionInfo {
```
sriumcp: `for _,` should be `for i,`

base/collect_database.go
```diff
@@ -0,0 +1,315 @@
...
+
+				// convert value to float
+				valueString := fmt.Sprint(value)
+				floatValue, err := strconv.ParseFloat(valueString, 64)
+				if err != nil {
+					log.Logger.Error("could not parse string \""+valueString+"\" to float:", err)
+					continue
+				}
+
+				err = exp.Result.Insights.updateMetric(metricName, mm, 0, floatValue)
```
sriumcp: 0 should be i

base/collect_database.go
```diff
@@ -0,0 +1,315 @@
...
+		if !info.IsDir() && strings.HasSuffix(path, "metrics.yaml") {
+			metricFilePaths = append(metricFilePaths, path)
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// collect metrics for all metric files and versionInfos
```
sriumcp, line 243: Before you enter this for loop, you may need to initialize results.

Please see https://github.com/iter8-tools/iter8/blob/800fe74b59559e99aeeb55ad4d2dc060472161a2/base/collect_grpc.go#L241-L244 for example.

sriumcp (changes requested):

base/collect_database.go
Alan-Cha: This is now being used properly in the tests.

base/collect_database.go
Alan-Cha: Added.

base/collect_database.go
Alan-Cha: My mistake!

base/collect_database.go
Alan-Cha: My mistake!

base/collect_database.go
Alan-Cha: Done!

base/collect_database.go
sriumcp: Could we please rename it as `type collectDatabaseTemplateInput` (note the small c), and also simply move the type definition to the test (since it is intended for use only in the test).

sriumcp (approved):
lgtm

Issue #1157, opened 2022-02-02
Helm based Kubernetes experimentation


sriumcp:
Evolve Iter8's Kubernetes experimentation capability to achieve the following.

1. Successive experiment launches for the same app: suppose an experiment is launched for app 'hello' (e.g., load testing with notifications/triggers at the end)... suppose a next version of `hello` comes along, and another experiment is launched. We do not want these experiments to clobber each other.

The simple solution which Helm provides is to simply push the new experiment into the cluster as the next version of a Helm release. This option is available to us now (since experiments are Helm templated, and we can `mixin` K8s templates into it). It would be tedious to implement this feature without Helm, but Helm already provides this capability.

2. The above feature is crucial for AutoX, which is the Iter8 controller for A/B(/n) as well as other experiment types that auto launches experiments.

3. Cluster agnostic: Different K8s clusters may be different ways of running jobs. A case in point being the "standard K8s cluster" which we've been experimenting with  using kind/Minikube, and non-standard ones like IBM cloud code engine, where jobs are not even supported.

We need to effectively isolate the core experiment objects (secret for spec, secret for result) from the deployment mechanism (job vs ce-job vs cron-job in the future). This is again something that can be achieved easily through the concept of mixins (i.e., not hardcoding the K8s deployment logic into Iter8, but allowing them to be specified independently of the experiment, example,  through another chart).

4. The following is a potential route to achieve both of the above ...
    ```shell
    iter8 hub --experiment-chart=load-test-http --runtime=k8s # can also be ce, or k8s-cron
    cd load-test-http
    iter8 k launch --app="hello" --set url="https://example.com" # this will create the next version of a Helm release
    ```

5. The current `iter8 k get`, `iter8 k assert`, `iter8 k run`, and `iter8 k report` functions are super good. We want their functionality to be retained as is (they all need testing; besides, if they can be simplified further in light of recent Helm based templating changes, then then need to be simplified).

6. Common flags supported by `iter8 k` needs to be retained. This is important functionality already supported by current `iter8 k`.

7. Revisit `iter8 k delete` in light of Helm based releases/templating.

sriumcp:
@kalantar FYI.

sriumcp:
Done.

PR #1189, opened 2022-03-15
Kubernetes experimentation

sriumcp:
Enables Kubernetes experimentation as described in #1157


README.md
```diff
@@ -16,45 +16,31 @@
...
+
+  > Grow your business with every release. Iter8 experiments can compare multiple versions based on business value and identify a winner.
+
+**Simple to use**
+
+  > Get started with Iter8 in seconds using pre-packaged experiment charts. Run Iter8 experiments locally, inside Kubernetes, or inside your CI/CD/GitOps pipelines.
+
+**App frameworks**
+
+  > Use with any app, serverless, or ML framework. Iter8 works with Kubernetes deployments, statefulsets, Knative services, KServe/Seldon ML deployments, or other custom Kubernetes resource types.
```
Alan-Cha:
```suggestion
  > Use with any app, serverless, or ML framework. Iter8 works with Kubernetes deployments, statefulSets, Knative services, KServe/Seldon ML deployments, or other custom Kubernetes resource types.
```

README.md
```diff
@@ -16,45 +16,31 @@
...
 ## :dart: Usage examples
 1.  [Load test, benchmark and validate HTTP services with SLOs](https://iter8.tools/0.9/tutorials/load-test-http/usage/).
 2.  [Load test, benchmark and validate gRPC services with SLOs](https://iter8.tools/0.9/tutorials/load-test-grpc/usage/).
-3.  Load test, benchmark and validate Knative services with SLOs: [HTTP](https://iter8.tools/0.9/tutorials/integrations/knative/load-test-http/) and [gRPC](https://iter8.tools/0.9/tutorials/integrations/knative/load-test-grpc/).
+3.  Performance testing and SLO validation using Iter8 GitHub Action: [HTTP](http://localhost:8000/latest/tutorials/load-test-http/ghaction/) and [gRPC](http://localhost:8000/latest/tutorials/load-test-grpc/ghaction/)
+4.  Performance testing and SLO validation for Knative services: [HTTP](https://iter8.tools/0.9/tutorials/integrations/knative/load-test-http/) and [gRPC](https://iter8.tools/0.9/tutorials/integrations/knative/load-test-grpc/).
 
-### Documentation
-Iter8 documentation is available at https://iter8.tools.
+Please see https://iter8.tools for complete documentation.
```
Alan-Cha:
```suggestion
Please see [https://iter8.tools](https://iter8.tools) for the complete documentation.
```

base/experiment.go
```diff
@@ -574,3 +583,219 @@ func (in *Insights) GetMetricsInfo(nm string) (*MetricMeta, error) {
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// NoFailure returns true if no task in the experiment has failed
+func (exp *Experiment) NoFailure() bool {
+	if exp != nil {
```
Alan-Cha, line 611:
```suggestion
return exp != nil && exp.Result != nil && !exp.Result.Failure
```

base/experiment.go
```diff
@@ -574,3 +583,219 @@ func (in *Insights) GetMetricsInfo(nm string) (*MetricMeta, error) {
...
+		if exp.Result != nil {
+			if !exp.Result.Failure {
+				return true
+			}
+		}
+	}
+	return false
```
Alan-Cha:
```suggestion
	return exp != nil && exp.Result != nil && !exp.Result.Failure
```

Alan-Cha (dismissed):

README.md
sriumcp: Is the extra space in front of `>` intentional?

sriumcp: It seems like I can commit your suggestions to my PR directly through the UI without having to do it in my local repo... hence the odd question. More context [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request).

Dockerfile
```diff
@@ -10,41 +10,43 @@ COPY ./ ./
...
-RUN curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
-RUN cp kustomize /bin
+# # Install Kustomize v3
+# RUN curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
+# RUN cp kustomize /bin
 
-# Install yq
-RUN GO111MODULE=on GOBIN=/bin go get github.com/mikefarah/yq/v4
+# # Install yq
+# RUN GO111MODULE=on GOBIN=/bin go get github.com/mikefarah/yq/v4
```
kalantar, line 35: Why are we making these changes?  These were here to support the exec action.

sriumcp: Good catch. As we introduce more capabilities in the charts, for example, notifications, we can reassess which of these binaries are actually required and put them back into the docker image.
***
For now, they are simply bloating the image, without actually contributing to the only two experiments we support through the charts today, namely `load-test-http` and `load-test-grpc`. Hence, the commenting out ...

kalantar:
I notice that the `hub`,`launch` and `k launch` commands no longer support the `--chartVersionConstraint` and `--repoURL` options. I think at least the `--repoURL` needs to be supported.

kalantar:
Out of curiosity, what are the `doc.go` files for?

sriumcp:
> I notice that the hub,launch and k launch commands no longer support the --chartVersionConstraint and --repoURL options. I think at least the --repoURL needs to be supported.

They are all still supported. Just hidden. We can unhide the flags if need be. In fact, these flags are used for testing.

> Out of curiosity, what are the doc.go files for?

Unambiguous package level documentation for multi-file go packages. See [here](https://github.com/natefinch/godocgo/blob/master/sub/doc.go) for example and also [here](https://go.dev/doc/effective_go) (under commentary).

kalantar:
>> I notice that the hub,launch and k launch commands no longer support the --chartVersionConstraint and --repoURL options. I think at least the --repoURL needs to be supported.

> They are all still supported. Just hidden. We can unhide the flags if need be. In fact, these flags are used for testing.

Ok. so I do see `--repoURL` defined in `NewHubOpts()`.  However, I do not see  `--chartVersionContraint` at all. Where is it?

sriumcp:
> However, I do not see --chartVersionContraint at all. Where is it?

[Here](https://github.com/sriumcp/iter8/blob/17338758f79f285b04289f8ef1aabf1ae3669ce6/action/hub.go#L26). Also see [here](https://github.com/helm/helm/blob/1ec0aacb8865d5b1f7ef1cb884bbf9b12579ecef/pkg/action/install.go#L119) and [here](https://github.com/sriumcp/iter8/blob/17338758f79f285b04289f8ef1aabf1ae3669ce6/cmd/root.go#L75).

sriumcp:
@kalantar Is this PR ready to merge or are there any comments / changes required?

kalantar (approved):
looks good.

Issue #1190, opened 2022-03-17
Migrate `readiness` task from v0.7


kalantar:
Example of proposed use:

```yaml
- task: k8s-objects-ready
  with:
    delay: 3 # delay in seconds before first check; optional; default 5s
    retries: 10 # number of attempts to check conditions; optional; default is 12
    retryInterval: 3 # time between attempts in seconds; optional; default is 5s
    objects: # list of k8s object references
    - kind: Pod # required
      name: curl-pod # required
      namespace: default # optional; default should determined from current context
      waitFor: condition=Ready # optional; see https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait
```

Implementation can be a port of what was used in v0.7.

For in cluster execution, user will need to have authorization to  get the objects. It may be necessary to create Role/RoleBinding. Is there a way to assist users to determine what these should be?

sriumcp:
`k8s-objects-ready` is nice! 
***
I suggest we have only a single object (instead of a list of objects) for the task to check. I also suggest removing the objects key, and simply inlining `kind`, `name`, `namespace`, and `waitFor` at the same level as delay and retries. This would make the experiment chart templates much easier to implement.

We can have multiple readiness checks in the experiment, simply by repeating the task.
***

kalantar:
How might we use this task with existing charts. For example, I imagine that I would want to combine one or more `k8s-objects-ready` tasks with the existing tasks in our current `load-test-http` chart to create a better load test. If I have a single task with a list of objects, I can imagine updating the `load-test-http` chart to optionally include it (depending on `experiment-config.yaml`. With multiple tasks, I do not see an easy way to combine. This increases the need to understand how to author charts.

kalantar:
> I suggest we have only a single object 
> ...
> We can have multiple readiness checks

This is a nice simplification

sriumcp:
Some usage examples I am thinking about ... (also documented in a separate k8s docs PR).
***
```shell
iter8 k launch -c load-test-http \
--set url=hello.svc.cluster.local \
--set ready.deploy.name=hello-there \
--set ready.svc.name=hello
```

The above would translate into the following experiment.yaml
```yaml
- task: k8s-objects-ready
  with:
    delay: 3 # delay in seconds before first check; optional; default 5s
    retries: 10 # number of attempts to check conditions; optional; default is 12
    retryInterval: 3 # time between attempts in seconds; optional; default is 5s
    kind: Deploy # required
[...]

PR #1191, opened 2022-03-17
Readiness

kalantar:
Initial implementation of readiness task.  See https://github.com/iter8-tools/iter8/issues/1190.

### DIFF
--- a/base/experiment.go
+++ b/base/experiment.go
@@ -153,6 +153,10 @@ func (s *ExperimentSpec) UnmarshalJSON(data []byte) error {
 		} else {
 			// this is some other task
 			switch *t.Task {
+			case ReadinessTaskName:
+				rt := &ReadinessTask{}
+				json.Unmarshal(tBytes, rt)
+				tsk = rt
 			case CollectHTTPTaskName:
 				cht := &collectHTTPTask{}
 				json.Unmarshal(tBytes, cht)
--- a/base/readiness.go
+++ b/base/readiness.go
@@ -1,0 +1,162 @@
+package base
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"regexp"
+	"time"
+
+	log "github.com/iter8-tools/iter8/base/log"
+)
+
+//TBD are RFC 1123 names sufficient?
+
+const (
+	// task name
+	ReadinessTaskName = "k8s-objects-ready"
+
+	// regex for resource names (RFC 1123 label names)
+	// see: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	labelFmt string = "[a-z0-9]([-a-z0-9]*[a-z0-9])?"
+
+	// maximum length of names
+	// see: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	maxLabelLength int = 63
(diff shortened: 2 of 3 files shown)


sriumcp:
The earlier `kubectl` based implementation of the readiness task had multiple issues... among other issues, we couldn't really unit test it (the previous attempt didn't actually have a unit test which tested the readiness logic; instead only peripheral stuff like name-length was unit tested).
***
Perhaps we can use the dynamic go client to `get` the resource in question...  like in these examples... 
***
Dynamic client example (List): https://itnext.io/generically-working-with-kubernetes-resources-in-go-53bce678f887
Another (Get): https://caiorcferreira.github.io/post/the-kubernetes-dynamic-client/
***
For checking conditions, perhaps we can marshal the unstructured object into [JSON ](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured.MarshalJSON), and unmarshal back to 
something like 

```go
type Resource struct {
  Status: Status
}
```

```go
type Status struct {
  Conditions: []Condition
}
```

where Condition is [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
```go
type Condition struct {
  Type               FooConditionType   `json:"type" description:"type of Foo condition"`
  Status             ConditionStatus    `json:"status" description:"status of the condition, one of True, False, Unknown"`

  // +optional
  Reason             *string            `json:"reason,omitempty" description:"one-word CamelCase reason for the condition's last transition"`
  // +optional
  Message            *string            `json:"message,omitempty" description:"human-readable message indicating details about last transition"`

  // +optional
  LastHeartbeatTime  *unversioned.Time  `json:"lastHeartbeatTime,omitempty" description:"last time we got an update on a given condition"`
  // +optional
  LastTransitionTime *unversioned.Time  `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
}
```

This should basically let us filter the "True" conditions, and check if the given condition type has status set to true.

```yaml
- task: k8s-object-ready # note: singular object as opposed to plural
  with:
    delay: 3 # delay in seconds before first check; optional; default 5s # optional with default
    retries: 10 # number of attempts to check conditions; optional; default is 12 # optional with default
    retryInterval: 3 # time between attempts in seconds; optional; default is 5s # optional with default
    kind: deployment # required
    apiVersion: v1 # required
    name: hello-there # required
    namespace: default # required
    conditionType: Ready # optional
```

kalantar:
Rewriting using unstructured APIs is doable. I've marked as draft while this is done. In the meantime this implementation allows us to play with the API and begin to think about what changes to the charts look like.

sriumcp:
Makes sense...

This PR is relevant to the chart design -- https://github.com/iter8-tools/hub/pull/10

Relevant part of the PR is here: https://github.com/sriumcp/hub/tree/k8s/testlibchart/load-test-http/templates

In particular, the `testlibcharts` folder shows how the experiment charts themselves can be drastically simplified (literally containing only 10 or 20 lines of helm template code or less), through the use of a library chart which provides commonly used templates.

kalantar:
closing; will reopen with version using k8s dynamic client

PR #1195, opened 2022-03-24
Readiness Task

kalantar:
Migration of readiness task from v0.7 to v0.9.  Uses the k8s dynamic client (replacing https://github.com/iter8-tools/iter8/pull/1191).  Related issue is https://github.com/iter8-tools/iter8/issues/1190.


base/kubedriver.go
```diff
@@ -0,0 +1,86 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+// KubeDriver embeds Kube configuration, and
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*EnvSettings
+	// Clientset enables interaction with a Kubernetes cluster
```
sriumcp: This is a typed client which the readiness task does not require. Please remove.

sriumcp (changes requested):
Initial change requested

base/kubedriver.go
kalantar: done. also another bit of repeated function.

kalantar:
Perhaps name of task should be singular `k8s-object-ready` instead of plural (`objects`).

base/kubedriver.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/rest"
+)
+
+// KubeDriver embeds Kube configuration, and
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*EnvSettings
```
sriumcp: Could we please embed *cli.EnvSettings (from Helm) instead of our own *EnvSettings? Specifically, we do not need our own struct for this. Like [this](https://github.com/iter8-tools/iter8/blob/802ad9fbc37dc0bdb0307b4dd169cc00e5e096a3/driver/kubedriver.go#L52).

The former provides a subset of what genericclioptions provides in terms of options, but it is more than enough (it includes the kubeconfig file; namespace; etc.).

Among other things, this should eliminate:
1. The need to have our own "EnvSettings" struct.2. 
2. We don't need the NewEnvSettings func.
3. We might be able to get rid of the `k8s.io/cli-runtime v0.23.4` (all our dependencies are very popular packages including k8s packages, but there are still high severity vulnerabilities in the security alerts; it would be nice to pay attention to the number of dependencies).
***

Also, if at all we need a struct which is in the "actual" code, we don't want that struct it in the test_helpers* file (because it is not just a test_helper anymore).

sriumcp (changes requested):
Another change requested.

base/kubedriver.go
kalantar: done

base/kubedriver.go
```diff
@@ -0,0 +1,68 @@
...
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*cli.EnvSettings
+	// RestConfig is REST configuration of a Kubernetes cluster
+	RestConfig *rest.Config
+	// DynamicClient enables unstructured interaction with a Kubernetes cluster
+	DynamicClient dynamic.Interface
+	// Namespace
+	Namespace *string
```
sriumcp: We can get rid of the `Namespace` field.
***
*cli.EnvSettings already has Namespace method which [does the right thing](https://github.com/helm/helm/blob/cba3b1eed4209bf1a6d052bf63adf7fbadb6af7d/pkg/cli/environment.go#L176).

So, instead of `kd.Namespace`, we can do `kd.Namespace()` everywhere.

base/kubedriver.go
```diff
@@ -0,0 +1,68 @@
...
+
+// KubeDriver embeds Kube configuration, and
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*cli.EnvSettings
+	// RestConfig is REST configuration of a Kubernetes cluster
+	RestConfig *rest.Config
+	// DynamicClient enables unstructured interaction with a Kubernetes cluster
+	DynamicClient dynamic.Interface
```
sriumcp: Could we please make this `dynamicClient` instead of `DynamicClient`?

base/kubedriver.go
```diff
@@ -0,0 +1,68 @@
...
+	"k8s.io/client-go/rest"
+)
+
+// KubeDriver embeds Kube configuration, and
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*cli.EnvSettings
+	// RestConfig is REST configuration of a Kubernetes cluster
+	RestConfig *rest.Config
```
sriumcp: It would be preferable to not have a `RestConfig` field in the struct, but instead simply have a `restConfig` local variable, wherever it is invoked (only in one place as far as I can see).

sriumcp: Among other things, it would enable us to directly embed `base.KubeDriver` into `driver.KubeDriver` (without any extra fields showing up in driver).

sriumcp (changes requested):
Please see changes requested.

base/test_helpers_driver.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"helm.sh/helm/v3/pkg/cli"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
+)
+
+// initKubeFake initialize the Kube clientset with a fake
+func (kd *KubeDriver) initKubeFake(objects ...runtime.Object) {
```
sriumcp: We should not use fake methods on real objects. Let us please make them functions instead. This is true even if the methods are not exposed externally.
***
If you look at the k8s APIs and Helm APIs, they both adhere to the above pattern.

sriumcp: Turns out I have made the same error as above while developing the `driver` package. If you wish, you can fix this in `driver`, or I will do so in my upcoming PR.

go.mod
```diff
@@ -39,6 +39,7 @@ require (
 	helm.sh/helm/v3 v3.8.1
 	k8s.io/api v0.23.5
 	k8s.io/apimachinery v0.23.5
+	k8s.io/cli-runtime v0.23.4
```
sriumcp: Where is this dependency being used? Could you please try `go mod tidy` and see if this goes away?

base/readiness_test.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nm,
+			Namespace: ns,
+		},
+	}
+	return (*podBuilder)(pod)
+}
+
+func (p *podBuilder) Build() *unstructured.Unstructured {
```
sriumcp: Could we please not externalize any test helper methods? like Build or WithCondition ... 
***
Limiting to package local scope is better (I realize podBuilder itself is package local; same idea).

base/driver.go
```diff
@@ -0,0 +1,9 @@
+package base
```
sriumcp: Could we please merge this one line file into base/kubedriver.go?

base/readiness.go
```diff
@@ -0,0 +1,197 @@
...
+	err := t.validateInputs()
+	if err != nil {
+		return err
+	}
+
+	// initialization
+	t.initializeDefaults()
+	kd.initKube()
+	// set Namespace (from context) if not already set
+	if t.With.Namespace == nil {
```
sriumcp: Move this into initializeDefaults()... 
```go
if t.With.Namespace == nil {
  t.With.Namespace = StringPointer(kd.Namespace())
}
```

sriumcp (changes requested):
Please see requested changes.

base/kubedriver.go
kalantar: done

base/kubedriver.go
kalantar: done

base/kubedriver.go
kalantar: done

go.mod
kalantar: done/ the dependency was removed in some of the cleanup

base/readiness_test.go
kalantar: done

base/driver.go
kalantar: done

base/readiness.go
kalantar: this depends on cluster initialization; moved that as well.

base/test_helpers_driver.go
kalantar: done including the `driver` package.

driver/test_helpers.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // initKubeFake initialize the Kube clientset with a fake
-func (kd *KubeDriver) initKubeFake(objects ...runtime.Object) {
+func initKubeFake(driver *KubeDriver, objects ...runtime.Object) {
```
sriumcp: Could we please name that argument `kd` instead of `driver`? It is easier to disambiguate between a `driver` variable, and the `driver` package in this manner...
***
This is simply a question of convenience. I realize that within the `driver` package, there is no name conflict.

base/readiness.go
```diff
@@ -0,0 +1,207 @@
...
+	// ReadinessTaskName is the task name
+	ReadinessTaskName = "k8s-object-ready"
+
+	// defaultTimeout is default timeout for readiness command
+	defaultTimeout = "10s"
+)
+
+// ReadinessInputs identifies the K8s object to test for existence and
+// the (optional) condition that should be tested (succeeds if true).
+type readinessInputs struct {
```
sriumcp, line 32: The task requires:
```go
Group: string
Version string
Resource: string
```

We are taking as inputs,
```go
APIVersion: string
Resource: string
```
and leaving out the group. 

And we're using a `gvr` function to parse out the `APIVersion` into `version` and `group`.

***

Could we please use the former, much more direct fields in the task? i.e., 
```go
Group: string
Version string
Resource: string
```

This will obviously simplify the task implementation. For example, it would get rid of the need for `gvr` function, as you will straight away use what is passed in as inputs to the task.

*** 
Perhaps less obvious, it would also simplify the task template definition and invocation. In particular, it would allow you to craft a template like ... 
```go
{{- define "task.ready" }}
- task: k8s-object-ready
  with:
{{ toYaml . | indent 4 }}
{{- end }}
```

and invoke it like this.

```go
{{- if .Values.service.name }}
{{ include "task.ready" .Values.service }}
{{- end }}
```

and 

```go
{{- if .Values.deploy.name }}
{{ include "task.ready" .Values.deploy }}
{{- end }}
```

sriumcp (changes requested):
Hopefully two last comments.

driver/test_helpers.go
kalantar: good point. i was also trying to distinguish it from the package global `kd`

sriumcp: Perhaps something other than `kd` and `driver` ... like `kDriver`, or `kubeDriver`?

base/readiness.go
kalantar: I made this change but I am not sure it does either of these things as cleanly as suggested. I may not understand the template approach.  But this is a separate discussion.

base/kubedriver.go
```diff
@@ -0,0 +1,61 @@
...
+// KubeDriver embeds Kube configuration, and
+// enables interaction with a Kubernetes cluster through Kube APIs
+type KubeDriver struct {
+	// EnvSettings provides generic Kubernetes options
+	*cli.EnvSettings
+	// dynamicClient enables unstructured interaction with a Kubernetes cluster
+	dynamicClient dynamic.Interface
+}
+
+type GetObjectFuncType func(*KubeDriver, *corev1.ObjectReference) (*unstructured.Unstructured, error)
```
sriumcp: This doesn't appear to be used anywhere. Should be removed?

sriumcp (changes requested):
One of the type defs seem unused.

base/kubedriver.go
kalantar: good catch

sriumcp (approved):
lgtm

Issue #1197, opened 2022-03-26
Simplify hosting of experiment charts


sriumcp:
**Is your feature request related to a problem? Please describe.**
The use of Helm charts for experiments brings a great deal of simplification by allowing users to `iter8 launch experiment-chart` with values. But hosting the experiment charts in Helm repos brings a fair amount of development complexity (not to mention, encouraging various hacks like the need to turn off verification checks in order for the Docker image to work). This aspect needs simplification.

**Describe the solution you'd like**
Work with a `charts` folder, instead of a `charts` repo. Among other things, the `charts` folder can be hosted as part of the `iter8` repo, removing the need for the `hub` folder. Of course, various Iter8 CLI commands need to provide options for using a local path to charts, removing the need to download the charts, and substantially easing local development and testing. The last two are cumbersome at the moment because of the round-tripping through the repo.

The above solution is close (but simpler) than what we had in `v0.8`.

`v0.9` also introduced chart versioning. We do to keep this aspect. In particular, the chart's minor version needs to match with Iter8 CLI's minor version always (should be checked in unit-tests).

**Additional context**
`iter8 launch` and `iter8 k launch` currently provide a single line mechanism for launching experiments (without the need  to invoke a `hub` command beforehand). We want to retain this simplicity of usage with the new changes.

Commands affected by this issue are:

```shell
iter8 hub
iter8 launch
iter8 k launch
```

Other `iter8` commands should remain untouched, though their test cases might need to change if the tests involve one or more of the affected commands above.

PR #1198, opened 2022-03-28
moving experiment charts into iter8

sriumcp:
This is a step towards #1197 

Once this PR is merged, the `charts` directory will be part of the `iter8` repo, and the `iter8 hub/launch/...` and other relevant commands may be subsequently modified to work with local charts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,5 +140,3 @@ iter8
 **/ghz-call-data.json
 **/ghz-call-data.bin
 **/ghz-call-metadata.json
-
-**/load-test-http/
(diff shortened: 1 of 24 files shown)


.gitignore
```diff
@@ -140,5 +140,3 @@ iter8
 **/ghz-call-data.json
 **/ghz-call-data.bin
 **/ghz-call-metadata.json
-
-**/load-test-http/
```
kalantar, line 144: why just this one? what about load-test-grpc?

sriumcp: The PR actually removes that line. Sometimes, when running tests, we pull down load-test-http and pollute other folders. We can selectively ignore load-test-http in those test folders if need be.

kalantar: I'm blind.

kalantar (approved):

PR #1210, opened 2022-04-05
Fix Hub and versioning

sriumcp:
1. The hub command should work even if we make local changes to code. This PR ensures this by fixing the Makefile, the hub and version commands.
2. Simplify Makefile. Metadata which used to be set earlier is now removed. We never used it up until now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ INSTALL_PATH ?= /usr/local/bin
 DIST_DIRS   := find * -type d -exec
 TARGETS     := darwin/amd64 linux/amd64 linux/386 windows/amd64
 BINNAME     ?= iter8
+ITER8_IMG ?= iter8/iter8:edge
 
 GOBIN         = $(shell go env GOBIN)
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -90,14 +83,6 @@ dist: build-cross
...
 	@rm -rf '$(BINDIR)' ./_dist
 
-.PHONY: docker-build
-docker-build:
-	docker build -f Dockerfile -t $(ITER8_IMG) .
-
-.PHONY: docker-push
-docker-push:
-	docker push $(ITER8_IMG)
-
```
kalantar: Why removing these commands? I know we are building our image with GH actions. However, as part of the development process I have often built my own image for testing. I find these make commands useful for this purpose.  It is true that the need has lessened significantly starting with v0.8 but I have done it several times since.

sriumcp: In this case, we will keep it.

kalantar (approved):

Issue #1212, opened 2022-04-06
Reduce number of places where there is a hardcoded version


kalantar:
The version is hard-coded in a number of places. Most common is in charts as `appVersion`. However, other occurrences exist as well. This makes updating versions more complex. Strive to minimize the number of places and consider a release time workflow to create a PR with the necessary changes to reduce chance of error.

PR #1217, opened 2022-04-10
Kubernetes experiments

sriumcp:
While attempting to run Kubernetes, I ran into a bunch of problems. This PR fixes some of them, and includes tests for them.

- [x] Created a post-release test for Kubernetes experimentation. This is post-release since we want to test with the Docker image created by the release process.
- [x] [A previous PR](#1206) removed payload / other URLs from HTTP and gRPC tasks in favor of local files. This PR enables them through charts (and curl commands). So payload URLs can be specified, without the need to worry about certificates etc. during load tests.
- [x] I had to remove the Docker build and push from Makefile. `ITER8_IMG ?= iter8/iter8:edge` is an invalid setting in makefile due to the presence of the colon. Additionally, using a colon in variable/filename is complicated as documented [here](https://stackoverflow.com/questions/2052406/escaping-colons-in-filenames-in-a-makefile). In particular, `make clean install` fails if I include that Docker build/push conf, and passes otherwise. Given that `docker build` and `push` can be accomplished through one-liners outside of the makefile, and given that we don't have a real need to build Docker images through CLIs, and given that it is starting to interfere with other commands in the makefile that are actually being used both during dev, and as part of CI, I think it makes sense to remove Docker build/push from makefile. Of course, if this is really required, we can enable it again in the future.
- [x] HTTP task now supports only a single version. This simplification makes every part of the code (iter8 CLI + charts) simpler. We can consider re-enabling multiple versions in the task if it ever becomes required during load tests in the future.
- [x] Removed appVersions which partly addresses #1212
- [x] Minor changes (like replacing the word subcommand with command in some places)
- [x] Previously, the owner of a result secret was the job that created it. Now it is the spec secret associated with it. They both have similar effect upon upgrades/deletions of experiments, however the code becomes simpler now.

***
**What remains to be done**

- [ ] Set `values.yaml` appropriately and also create proper documentation with `values.yaml`. TBD after the merge, since we will use the task  links within values.yaml docs to describe parameters.

### DIFF
--- a/charts/iter8lib/templates/_task-grpc.tpl
+++ b/charts/iter8lib/templates/_task-grpc.tpl
@@ -40,6 +40,6 @@
 # task: generate gRPC requests for app
 # collect Iter8's built-in gRPC latency and error-related metrics
 - task: gen-load-and-collect-metrics-grpc
-with:
+  with:
 {{ toYaml $vals | indent 4 }}
 {{- end }}
--- a/charts/iter8lib/templates/_task-grpc.tpl
+++ b/charts/iter8lib/templates/_task-grpc.tpl
@@ -1,12 +1,12 @@
 {{- define "task.grpc" -}}
-{{- /* Validate values */ -}}
+{{/* Validate values */}}
 {{- if not .Values.host }}
 {{- fail "Please set a value for the host parameter." }}
 {{- end }}
 {{- if not .Values.call }}
 {{- fail "Please set a value for the call parameter." }}
 {{- end }}
-{{- /* Perform the various setup steps before the main task */ -}}
+{{/* Perform the various setup steps before the main task */}}
 {{- $vals := mustDeepCopy .Values }}
 {{- if .Values.protoURL }}
 # task: download proto file from URL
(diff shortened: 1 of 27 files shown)


Alan-Cha (approved):

charts/iter8lib/templates/_task-grpc.tpl
```diff
@@ -1,7 +1,45 @@
...
+# task: download proto file from URL
+- run: |
+    curl -o ghz.proto {{ $vals.protoURL }}
+{{- $pf := dict "proto" "ghz.proto" }}
+{{- $vals = mustMerge $pf .Values }}
+{{- end }}
+{{- if .Values.dataURL }}
+# task: download JSON data file from URL
+- run: |
+    curl -o data.json {{ $vals.dataURL }}
```
kalantar, line 21: It is my impression that the purpose of this (and similar curl commands) here and in `_task-http.tpl` is to avoid worrying about certificates. I am not sure I see how using curl avoids this problem. To be more bulletproof, we might include `-k` and `-L` as options.

kalantar (approved):

PR #1218, opened 2022-04-11
readiness chart changes

kalantar:
Chart changes to support addition of readiness task.  Addresses issue https://github.com/iter8-tools/iter8/issues/1222


kalantar:
The current task template supports identifying a single resource for each kind to check for readiness by specifying: 

```shell
--set ready.<kind>=a name
```

If using `--set ready.timeout=timeout` to specify a timeout or `--set ready.namespace=ns` to specify a namespace, these properties will apply to all resources.

charts/load-test-http/values.yaml
```diff
@@ -28,3 +28,20 @@
 ###   http/latency-mean: 50
 ###   http/latency-p99: 200
 # SLOs: null
+
+### readinessDefinitions   A map of Kubernetes resource kinds to group/version/resource and condition
+### group/version/kind and namespace/name can be used to uniquely identify a resource in a cluster
+### condition is a status condition that indicates the resource is ready for use
+readinessDefinitions:
+  pod:
```
sriumcp: `pod` is not something that any of use cases directly involve (deployment; knative; kserve, Seldon, statefulset -- all create pods, but we don't check for the pod readiness in any of these cases).
***
Let us please remove `pod`.

.github/workflows/tests.yaml
```diff
@@ -123,6 +123,45 @@ jobs:
...
+        docker pull kennethreitz/httpbin
+        docker run -p 80:80 kennethreitz/httpbin &
+        HOST_IP=$(ip -f inet addr show docker0 | grep -Po 'inet \K[\d.]+')
+        echo "HOST_IP=$HOST_IP" >> $GITHUB_ENV
+        while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://$HOST_IP/get)" != "200" ]]; do
+          sleep 5; 
+        done;
+    - name: launch load-test-http with readiness checks
+      run: |
+        iter8 launch -c load-test-http --noDownload \
```
sriumcp, line 201: This  invocation is exactly what we want.

charts/load-test-grpc/templates/_experiment.tpl
```diff
@@ -1,4 +1,5 @@
 {{ define "experiment" -}}
+{{- include "task.readiness" . }}
```
sriumcp: looks good.

sriumcp (changes requested):
Initial fixes requested.

charts/load-test-http/values.yaml
kalantar: Agreed. Pod is not very interesting. Are there other resources we would like to add to this list now?

sriumcp: For now, service and deployment seem good.

charts/load-test-grpc/values.yaml
```diff
@@ -41,3 +41,21 @@
 ###   grpc/latency/mean: 50
 ###   grpc/latency/p99: 200
 # SLOs: null
+SLOs: null
+
+### readinessDefinitions   A map of Kubernetes resource kinds to group/version/resource and condition
+### group/version/kind and namespace/name can be used to uniquely identify a resource in a cluster
+### condition is a status condition that indicates the resource is ready for use
+readinessDefinitions:
```
sriumcp: The readiness definitions are essentially constants, as opposed to other values, like deployment name.
***
Could we please move these definitions under `iter8lib` chart's `values.yaml`, especially, under its global values?

```yaml
# iter8lib/values.yaml
global:
  readinessDefinitions:
    deploy: ...
```

It can then be set/accessed as `iter8lib.global.readinessDefinitions` in every other chart that embeds it.

sriumcp (changes requested):
One more change requested.

charts/load-test-http/values.yaml
kalantar: removed

charts/load-test-grpc/values.yaml
kalantar: done

charts/iter8lib/templates/_task-readiness.tpl
```diff
@@ -0,0 +1,32 @@
+{{- define "task.readiness" -}}
```
sriumcp: I want to propose a similar template, but simpler and easier to read, and as it happens, without any need for values in `iter8lib`.
***
```yaml 
# in a separate file
{{- define "task.ready.tn" }}
{{- if .Values.ready.timeout }}
timeout: {{ .Values.ready.timeout }}
{{- end }}
# .Values.ready.namespace useful only in local experiments
{{ $namespace := "" }}
{{- if .Values.ready.namespace }}
{{ $namespace = .Values.ready.namespace }}
{{- end }}
# .Release.Namespace useful only in Kubernetes experiments
{{- if .Release.Namespace }}
{{ $namespace = .Release.Namespace }}
{{- end }}
{{-if $namespace }}
{{ namespace: $namespace }}
{{- end }}
{{- end }}
```

```yaml
{{- define "task.ready" }}
{{- if .Values.ready.service }}
- task: k8s-object-ready
   with:
    name: {{ .Values.ready.service }}
    version: v1
    resource: services
{{- include "task.ready.tn" . | indent 4 }}
{{- end }}
{{- if .Values.ready.deploy }}
- task: k8s-object-ready
   with:
    name: {{ .Values.ready.deploy }}
    version: v1
    resource: deployments
    condition: Available
{{- include "task.ready.tn" . | indent 4 }}
{{- end }}
{{- end }}
```

sriumcp (changes requested):

charts/iter8lib/templates/_task-readiness.tpl
kalantar: done

sriumcp (approved):
lgtm

PR #1220, opened 2022-04-12
fixes to get percentiles working

### LINKED ISSUE #1219: Percentiles not working properly in http load test
**Describe the bug**
Setting non-default SLO percentiles, like `97.5` used to work, but not anymore.

**To Reproduce**
Try the Iter8 load test tutorial (v0.10) this way...

```
iter8 launch -c load-test-http --noDownload --set url=http://127.0.0.1/get --set SLOs.http/error-rate=0 --set SLOs.http/latency-mean=50 --set SLOs.http/latency-p90=100 --set SLOs.http/latency-p'97\.5'=200
```

Followed by `iter8 report`

**Expected behavior**
User should be able to set any percentile value in SLOs without errors.

sriumcp:
Fixes #1219

Also includes ... 

- [x] Other bug fixes in http and gRPC task templates
- [x] Minor changes to README.md, and CONTRIBUTING.md
- [x] Adding values.yaml documentation
- [x] Adding doc.go documentation

### DIFF
--- a/base/collect_http.go
+++ b/base/collect_http.go
@@ -21,8 +21,8 @@ type errorRange struct {
 	Upper *int `json:"upper" yaml:"upper"`
 }
 
-// collectHTTPInputs contain the inputs to the metrics collection task to be executed.
-type collectHTTPInputs struct {
+// CollectHTTPInputs contain the inputs to the metrics collection task to be executed.
+type CollectHTTPInputs struct {
 	// NumRequests is the number of requests to be sent to the app. Default value is 100.
 	NumRequests *int64 `json:"numRequests" yaml:"numRequests"`
 	// Duration of this task. Specified in the Go duration string format (example, 5s). If both duration and numQueries are specified, then duration is ignored.
@@ -103,7 +103,7 @@ type collectHTTPTask struct {
 	// TaskMeta has fields common to all tasks
 	TaskMeta
 	// With contains the inputs to this task
-	With collectHTTPInputs `json:"with" yaml:"with"`
+	With CollectHTTPInputs `json:"with" yaml:"with"`
 }
 
 // initializeDefaults sets default values for the collect task
--- a/base/collect_http.go
+++ b/base/collect_http.go
@@ -21,8 +21,8 @@ type errorRange struct {
 	Upper *int `json:"upper" yaml:"upper"`
 }
 
-// CollectHTTPInputs contain the inputs to the metrics collection task to be executed.
-type CollectHTTPInputs struct {
+// collectHTTPInputs contain the inputs to the metrics collection task to be executed.
+type collectHTTPInputs struct {
 	// NumRequests is the number of requests to be sent to the app. Default value is 100.
 	NumRequests *int64 `json:"numRequests" yaml:"numRequests"`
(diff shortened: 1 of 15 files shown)


charts/load-test-http/values.yaml
```diff
@@ -1,54 +1,20 @@
...
-
-### numQueries    Number of requests sent to the app. 
-# numQueries: null
-
-### duration    Duration for which requests are sent to the app.
-### Value can be any Go duration string (https://pkg.go.dev/maze.io/x/duration#ParseDuration)
-### This field is ignored if `numQueries` is specified.
-# duration: null
+### This chart accepts all parameters that are part of the collectHTTPInputs struct.
+### https://github.com/iter8-tools/iter8/blob/2891c13aac668bd651cc8e76f2d11a9d9a4fb75b/base/collect_http.go#L25
```
kalantar, line 6: The type has been capitalized. It might be better not to include a line number om the link since this can change. And perhaps reference to master branch instead of a particular commit.

base/collect_http.go
```diff
@@ -21,8 +21,8 @@ type errorRange struct {
 	Upper *int `json:"upper" yaml:"upper"`
 }
 
-// collectHTTPInputs contain the inputs to the metrics collection task to be executed.
-type collectHTTPInputs struct {
+// CollectHTTPInputs contain the inputs to the metrics collection task to be executed.
```
kalantar: I do not understand why this has been capitalized. I do not see its  use outside of the package. The only code changes in this PR are to the base package.

sriumcp: The only reason for capitalizing it is show that its documentation shows up in go.pkg.dev, which in turn I can reference in the values.yaml file.
***
Considering we can use the GitHub permalinks instead, I also don't see the value in  capitalizing the type. Reverting ...

charts/load-test-http/values.yaml
sriumcp: This is the permalink Iter8 gives, so unless, the repo is removed, I believe the link will always be valid and refer to the same thing, which is the effect we  want?
***
Once we created a tagged release, I can follow up with another PR to update the perma link.

kalantar: I guess if we want to keep updating this link with each release we can do that. The link remains valid since it points to a particular commit. However, the code might change after this commit and we want the documentation to point at the recent version. I know. It is version specific. But I am thinking about maintenance here. I'm willing to trade off what someone looking at a particular version might see.

base/collect_http.go
kalantar: I was going to say this seems like a good reason :)  This never occurred to me.

charts/load-test-http/values.yaml
sriumcp: The only reason to change the permalink would be that the underlying struct (and hence the supported parameters/default values) have changed.
***
Until then, they can look at a fixed link given in the values.yaml file, even if it is not the latest.

sriumcp: The idea is *not* to change on every release. But to change on every change to the struct/values.yaml.

kalantar (approved):

Issue #1223, opened 2022-04-14
Introduce GitHub action to install Iter8


kalantar:
The existing Iter8 action uses a docker implementation and calls `iter8 run`, `iter8 assert` and `iter8 report`
Changes to Iter8 mean that the action becomes invalid. To make it more resilient and to simplify the documentation, rewrite the action to just install Iter8. The user can implement the additional steps with a few lines.

PR #1224, opened 2022-04-18
GitHub action

kalantar:
Add action to install Iter8. Addresses https://github.com/iter8-tools/iter8/issues/1223


kalantar:
Identifying the default version to install should depend on the version of the action being used rather than hard-codimg it.

kalantar:
This is ready for review. The action installs the latest release of Iter8. It takes an optional parameter `version` that overrides the version to be installed.

action.yaml
```diff
@@ -0,0 +1,29 @@
+name: 'Iter8 Install'
+description: 'Install Iter8 CLI'
+inputs:
+  version:
```
sriumcp: Could we please get rid of the `version` input?
***
```yaml
- uses: iter8-tools/iter8@v0.10.5
```

or 

```
- uses: iter8-tools/iter8@master
```
are the two types of usage we want to support. In the former case, the version info is explicitly supplied. In the latter case, we want to use your logic for figuring out the latest release, and install it. In both cases, you can extract the [`github.action_ref`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), and do not required a version to be supplied as a separate input.

action.yaml
```diff
@@ -0,0 +1,29 @@
...
+    description: "Version of iter8 to install"
+    required: false
+runs:
+  using: 'composite'
+  steps:
+    - name: Install Go
```
sriumcp: What is the need for this step?

action.yaml
```diff
@@ -0,0 +1,29 @@
...
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+    - name: Install latest version of Iter8 CLI
+      run: | 
+        OS=$(echo ${{ runner.os }} | tr '[:upper:]' '[:lower:]')
+        ARCH=$(echo ${{ runner.arch }} | sed 's/X86/386/' | sed 's/X64/amd64/')
```
sriumcp, line 18: What happens if `runner.os` or `runner.arch` values happen to be  values for  which we do not produce binary assets?
***
I confess I don't know bash well enough to understand the `OS` and `ARCH` definitions.

sriumcp (changes requested):

action.yaml
kalantar: good catch. Initially I used `go install`

action.yaml
kalantar: This is a possibility. The install will fail (no binary will be found). This can be modified to explicitly check.

Sadly, the values for os and arch do not match the values in the binaries. Similarly if I use alternatives like `uname`. Some sort of translation is needed. The bash does these translations (for os just convert to lower case; for arch change it appropriately)

action.yaml
kalantar: done.

action.yaml
kalantar: This is better documented now. The action does better error checking on download to provide a better explanation of any failure. But no assumption is made of what are valid values so that no changes should be necessary if new artifacts are added in the future.

action.yaml
kalantar: The context variable `github.action_ref` is always the empty string in the `run`. https://github.community/t/getting-information-about-my-action/231562 identifies a work around by reading the context variable in the step `env`. In this way we are able to implement this.

In particular, 

- `@vMAJAOR.MINOR.PATCH` will use the matching version
- `@vMAJOR.MINOR` will identify the latest release matching MAJOR and MINOR -- assuming we create at tags `vMAJOR.MINOR`
- `@stable` will identify the latest release (is this a suitable release name? feel that `@master` or `@latest` will cause misunderstandings)

kalantar:
Currently, the test case for action will load the latest released version of Iter8, not the version in the current build.

sriumcp (approved):
lgtm

PR #1225, opened 2022-04-20
Istio SLO experiment

Alan-Cha:


base/collect_database.go
```diff
@@ -115,6 +114,7 @@ func getElapsedTime(versionInfo map[string]interface{}, exp *Experiment) (int64,
 		case int64:
 			startingTime = rawStartingTime
 		case int:
+		case float64: // parsing the metrics file gives float64
```
Alan-Cha, line 117: Perhaps this can be resolved by adding "| int" in the metrics template

charts/iter8lib/templates-temp/_k-manifest-cronjob.tpl
```diff
@@ -0,0 +1,13 @@
+{{- define "k.manifest-cronjob" -}}
+{{ include "k.cronjob" . }}
+---
+{{ include "k.spec.secret-istio" . }}
```
Alan-Cha: I am still resorting to an Istio specific secret. Perhaps we can discuss how to avoid this?

charts/iter8lib/templates/_k-manifest-istio.tpl
```diff
@@ -0,0 +1,13 @@
+{{- define "k.manifest-istio" -}}
+{{ include "k.job" . }}
+---
+{{ include "k.spec.secret-istio" . }}
```
Alan-Cha, line 4: I am still resorting to an Istio specific secret. Perhaps we can discuss how to avoid this?

Alan-Cha:
I did not fix the `reporter='source'` in the metrics file yet because I was thinking, this is not so much a mirroring experiment. It's more of a generic SLO test, that can be used for both the V1 and the V2 whether there is mirroring or not. Iter8 doesn't know when which version it is working with and whether there is mirroring or not. That means, it's ultimately up to the user to decide whether `reporter='source'` should be configured. However, there seems like kind or an odd and technical option to provide to the user. I want to discuss it first before moving forward.

Alan-Cha:
The failing tests seem to be related to the unit tests I created for the database task. I think the issue has to do with differences in URL encoding between different versions and different packages. I will look into an alternative way of testing. In the mean time, ignore the unit tests.

Alan-Cha:
Note that I moved the Istio metrics template from `iter8lib` into the `istio-slo-validation`. I think it is possible to move it back into `iter8lib` but I need to do some refactoring.

Alan-Cha:
You can try it out by following these steps

Set up Minikube
https://istio.io/latest/docs/setup/platform-setup/minikube/

Install Istio
```
istioctl install
```

Install Prometheus add on
https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-quick-start

Follow the mirroring tutorial
https://istio.io/latest/docs/tasks/traffic-management/mirroring/

Send a few extra queries
```
kubectl exec "${SLEEP_POD}" -c sleep -- curl -sS http://httpbin:8000/headers
```

In another terminal, run
```
minikube tunnel
```

In a third terminal, run
```
istioctl dashboard prometheus
```

Then run

```
iter8 launch -c istio-slo-validation --set endpoint=http://127.0.0.1:9090/graph --set destination_workload=httpbin-v2 --set destination_workload_namespace=default --set StartingTime=1647360597 --noDownload
```

sriumcp:
> However, there seems like kind or an odd and technical option to provide to the user. I want to discuss it first before moving forward.

One possibility is to make this an option, but to set a default (in the case of the mirroring experiment, the default for `reporter` will be `destination`).

sriumcp:
> Note that I moved the Istio metrics template from `iter8lib` into the `istio-slo-validation`. I think it is possible to move it back into `iter8lib` but I need to do some refactoring.

I think it  is ok to start with the metrics template in `istio-slo-validation`. Once we get a handle on the initial version of the experiment, we can consider moving it into `iter8lib`.

action/gen.go
```diff
@@ -63,6 +104,21 @@ func (gen *GenOpts) LocalRun() error {
 		Data: eData,
 	})
 
+	// check for a metrics template and get its source
```
sriumcp: I feel our design for reading in `*.metrics.tpl` and creating `*.metrics.yaml` in `gen`, and then reading `*.metrics.yaml` again within the task, is a bit hairy, and whole lot more complex than it needs to be.
***
Do you think it makes sense to simplify this logic by simply supplying a list of providers to the experiment task?

```yaml
- task: collect-metrics-database
  with:
    providers:
    - istio
    - knative
    versionInfo: ...
```

The above design will allow you to specifically read `istio.metrics.yaml`, and `knative.metrics.yaml` within the `collect-metrics-database` task.

You can factor out `providers` as a value, and supply a default value in the chart (`[istio]` is the default value for the istio-slo-validation chart of course).

sriumcp (changes requested):

action/gen.go
```diff
@@ -63,6 +64,33 @@ func (gen *GenOpts) LocalRun() error {
...
 	})
 
+	/*
+		attempt to extract providers from valuesToRender
+
+		if providers are available, create the respective metrics files from the
+		templates
+	*/
+	var providers []string
+	if rawProviders, ok := c.Values["providers"]; ok {
```
kalantar, line 74: Please define a use a constant instead of hardcoding strings. This simplifies inconsistencies and simplifies changes.

charts/istio-slo-validation/templates/_experiment.tpl
```diff
@@ -0,0 +1,4 @@
+{{ define "experiment" -}}
```
kalantar, line 1: Should we include the readiness check step as well?

```
{{- include "task.ready" . -}}
```

charts/istio-slo-validation/values.yaml
```diff
@@ -0,0 +1,70 @@
+# Prometheus URL
```
kalantar, line 1: I am surprised to see `Endpoint` and `istio/error-rate` here. I would have expected these to be empty and expect the user to fill in. Endpoint probably as required.

---

Would lower case `endpoint` be better?

charts/iter8lib/templates/_k-manifest-istio.tpl
```diff
@@ -0,0 +1,13 @@
...
+---
+{{ include "k.spec.role" . }}
+---
+{{ include "k.spec.rolebinding" . }}
+---
+{{ include "k.result.role" . }}
+---
+{{ include "k.result.rolebinding" . }}
+{{- end }}
```
kalantar, line 13: I am a bit bothered by the istio specific items in `iter8lib`. I wonder if it makes more sense to create a `istiolib` or `iter8istio` that contains these. The dependency might be `istio-slo-validation` depends on `istiolib` depends on `iter8lib`.

kalantar: Should be updated to reflect new changes to `_k-manifest.tpl` (ie, conditionals and readiness).

charts/iter8lib/templates/_k-spec-secret-istio.tpl
```diff
@@ -0,0 +1,12 @@
...
+{{- $name := printf "%v-%v" .Release.Name .Release.Revision -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $name }}-spec
+stringData:
+  experiment.yaml: |
+{{ include "experiment" . | indent 4 }}
+  metrics.yaml: |
+{{ include "istio.metrics" . }}
```
kalantar, line 11: `| indent 4` ?

charts/iter8lib/templates/_k-spec-secret-istio.tpl
```diff
@@ -0,0 +1,12 @@
+{{- define "k.spec.secret-istio" -}}
+{{- $name := printf "%v-%v" .Release.Name .Release.Revision -}}
```
kalantar, line 2: @sriumcp perhaps you can comment on the elements that are essential from `_k-spec-secret.tpl`. `name` and `annotations`?

charts/istio-slo-validation/templates/_experiment.tpl
sriumcp: This is an interesting point... 

Ideally, you would want wait until the shadow deployment is ready before you start mirroring. But the task really wouldn't accomplish this since the experiment is totally decoupled from the mirroring mechanism...

I am inclined to think this experiment is not an intended use case for readiness check (within experiment), so we can leave this task out of the experiment.

charts/istio-slo-validation/values.yaml
sriumcp: +1 for lower case `endpoint`

charts/iter8lib/templates/_metrics-ce.tpl
```diff
@@ -0,0 +1,83 @@
+{{- define "metrics.ce" -}}
+# endpoint where the monitoring instance is available
+# https://cloud.ibm.com/docs/monitoring?topic=monitoring-endpoints#endpoints_sysdig
+url: {{ .Values.Endpoint }}/prometheus/api/v1/query # e.g. https://ca-tor.monitoring.cloud.ibm.com
```
kalantar, line 4: Does CE use Prometheus or sysdig? It says prometheus here but `provider` below refers to sysdig

charts/iter8lib/templates/_metrics-common-istio.tpl
```diff
@@ -0,0 +1,116 @@
+{{- define "metrics.common.istio" }}
```
kalantar, line 1: I am confused by the need for this. We seem to be extracting the variables from the dictionary `,Values` and added them to the dictionary `.`. We are then using them in templates directly. Why do we need to do this?

charts/iter8lib/templates/_k-spec-secret-istio.tpl
Alan-Cha: This is used to indent the resolved value, yes? I see this pattern in other places in `iter8lib` such as [here](https://github.com/iter8-tools/iter8/blob/cab1e42dd5192d757d9261e4eab4a113654345bb/charts/iter8lib/templates/_task.ready.tpl#L34). If you think I should remove them, then I will.

action/gen.go
Alan-Cha: I'm not really sure when we should use constants vs hardcoded strings. I try to use constants if there is a string that I need to use multiple times. To clarify, this is not one of those instances.

Should we use constants for all strings? Or whenever we reuse strings? Or whenever we need to access a struct field?

charts/iter8lib/templates/_metrics-common-istio.tpl
Alan-Cha: This template is used to reduce replication in the metrics template. For example it is used 8 times in the Istio metrics template.

https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L65
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L86
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L107
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L121
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L145
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L160
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L183
https://github.com/iter8-tools/iter8/blob/963bb43f3dcf6580fd2fc9e1ea6d3066cef67fa7/charts/iter8lib/templates/_metrics-istio.tpl#L197

Let me know if I am misunderstanding your question.

charts/iter8lib/templates/_metrics-ce.tpl
Alan-Cha: CE uses both. I can change this. I think this is just a remnant of my early explorations.

kalantar:
@Alan-Cha is there any reason to keep this open?

PR #1228, opened 2022-04-22
Fixing Kubernetes experimentation

sriumcp:
Fixing bugs in `k launch` command.
Updating various command documentation to make them correct/accurate.

### DIFF
--- a/cmd/hub.go
+++ b/cmd/hub.go
@@ -8,9 +8,9 @@ import (
 
 // hubDesc is the description of the hub command
 const hubDesc = `
-Download an experiment chart to a local directory.
+Download Iter8 experiment charts from a Git repo.
 
-	$ iter8 hub -c load-test-http
+	$ iter8 hub
 
 This command is intended for development and testing of experiment charts. For production usage, the iter8 launch command is recommended.
 `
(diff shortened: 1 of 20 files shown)


cmd/hub.go
```diff
@@ -8,9 +8,9 @@ import (
 
 // hubDesc is the description of the hub command
 const hubDesc = `
-Download an experiment chart to a local directory.
+Download Iter8 experiment charts from a Git repo.
```
kalantar: Are we restricted to a Git repo? My impression from the code is that it could be from any of many places (there is a list of getters used by Get()).

cmd/klaunch.go
```diff
@@ -15,17 +15,17 @@ Launch an experiment in Kubernetes.
 
 	$ iter8 k launch -c load-test-http --set url=https://httpbin.org/get
 
-To locally render the Kubernetes experiment, use the dry option.
+Use the dry option to simulate a Kubernetes experiment. This creates the manifest.yaml file.
```
kalantar: Perhaps add the phrase " but does not deploy it." to be more explicit?

sriumcp: Fixed this documentation.

kalantar (approved):

PR #1229, opened 2022-04-25
Fixing Kubernetes experimentation bugs

sriumcp:
1. Creating a secret also requires waiting for permissions. 
2. Also increasing the timeouts for readiness [due to these failed tests](https://github.com/iter8-tools/iter8/runs/6158808866?check_suite_focus=true).
3. Improving variable/flag naming along with documentation for remote Iter8 charts URL.

### DIFF
--- a/driver/kubedriver.go
+++ b/driver/kubedriver.go
@@ -281,20 +281,32 @@ func (driver *KubeDriver) formResultSecret(r *base.ExperimentResult) (*corev1.Se
 // createExperimentResultSecret creates the experiment result secret
 func (driver *KubeDriver) createExperimentResultSecret(r *base.ExperimentResult) error {
 	if sec, err := driver.formResultSecret(r); err == nil {
-		secretsClient := driver.Clientset.CoreV1().Secrets(driver.Namespace())
-		_, err1 := secretsClient.Create(context.Background(), sec, metav1.CreateOptions{})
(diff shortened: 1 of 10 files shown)


driver/kubedriver.go
```diff
@@ -281,20 +281,32 @@ func (driver *KubeDriver) formResultSecret(r *base.ExperimentResult) (*corev1.Se
...
+				secretsClient := driver.Clientset.CoreV1().Secrets(driver.Namespace())
+				_, err3 := secretsClient.Create(context.Background(), sec, metav1.CreateOptions{})
+				return err3
+			},
+		)
 		if err1 != nil {
-			err2 := fmt.Errorf("unable to create secret %v", sec.Name)
-			log.Logger.WithStackTrace(err1.Error()).Error(err2)
-			return err2
+			err4 := fmt.Errorf("unable to create secret %v", sec.Name)
```
Alan-Cha, line 302: Is it good practice to enumerate error variables like this? Seeing it go to 4 makes me more and more concerned. Other than that, everything else looks good.

Alan-Cha (dismissed):

driver/kubedriver.go
sriumcp: I started by reusing error variable names ... until it started leading to subtle bugs. Given that readability is not impacted in anyway, I feel this is alright.

kalantar (approved):

PR #1230, opened 2022-04-25
Fix more bugs in Kubernetes experimentation

sriumcp:
1. It turns out that while specificying RBACs for create verbs, one cannot specify resource names ... please refer to the note [here](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources). This leaves us with two options ... 
  - create the experiment spec and result as part of the Helm install, and the experiment job only does update of the created secret.
  - create the result in the experiment job, but make the create RBAC expansive (i.e., create secrets, instead of creating secret with a specific name).

I chose the first option to keep the RBAC profile minimal. For starters, I don't need a create verb in the RBAC at all. I also eliminated release names from Helm-installed-resource-names. This means, the RBACs do not have to change from release-to-release, paving the way eventually for autox.


.github/workflows/assets.yaml
```diff
@@ -98,18 +98,13 @@ jobs:
     runs-on: ubuntu-latest
     needs: build-and-push
     steps:
-    - name: Install Iter8
-      run: |
-        tagref=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
-        wget https://github.com/iter8-tools/iter8/releases/download/$tagref/iter8-linux-amd64.tar.gz
-        tar -xvf iter8-linux-amd64.tar.gz
-        mv linux-amd64/iter8 /usr/local/bin
+    - uses: iter8-tools/iter8@v0.10
```
kalantar, line 101: This won't work until we create a v0.10 branch or tag. Did we already do this?

charts/iter8lib/templates/_k-result-secret.tpl
```diff
@@ -0,0 +1,14 @@
+{{- define "k.result.secret" -}}
```
kalantar, line 1: I am still wondering about the implications of a single spec and single result secret. This means there is no history of experiments. In practice, only 1 experiment per "group". Is this acceptable? I suppose it the interest of keeping it simple, it might be. Does this impact `delete`?

sriumcp: > In practice, only 1 experiment per "group". Is this acceptable?

For now, we definitely want to keep things simple and be able to make progress. This approach simplifies not only `iter8 k` based experimentation, but also simplifies `autox` which we will want to get into later on.

Incidentally, the previous implementation also did not retain history ... it merely named the experiment secrets uniquely with revision numbers, which has the effect of complicating RBACs (different RBACs in each experiment revision).

This does not change `delete`, as delete was always intended to wipe out the entire group (i.e., the underlying Helm release), which it still does.

***

Obviously, if maintaining history becomes important for some new use-case, we can always dump the results of experiments (including Kubernetes experiments someplace, and view it and create reports out of it offline). This possibility still exists.

.github/workflows/assets.yaml
sriumcp: Good catch. Done now.

kalantar (approved):

PR #1232, opened 2022-04-26
Fix Kubernetes experimentation

sriumcp:
Fix more issues in Kubernetes experimentation
1. Revisions in annotations are now strings
2. Newer version of kind is required for `kubectl wait` to work
3. Format of experiment.yaml in secret is fixed
4. Job manifest can be selectively disabled (useful for testing purposes)


kalantar (dismissed):
lgtm

PR #1249, opened 2022-05-02
Db task with providers

Alan-Cha:
The db task now looks for metrics files based on the `providers` rather than using all files that ends with `.metrics.yaml` in the current directory.

### DIFF
--- a/base/collect_database.go
+++ b/base/collect_database.go
@@ -46,9 +46,9 @@ type Params struct {
 }
 
 const (
-	StartingTime = "StartingTime"
-	ElapsedTime  = "ElapsedTime"
-	TimeLayout   = "Jan 2, 2006 at 3:04pm (MST)"
+	startingTimeStr = "StartingTime"
+	elapsedTimeStr  = "ElapsedTime"
+	timeLayout      = "Jan 2, 2006 at 3:04pm (MST)"
 )
 
 type collectDatabaseInputs struct {
@@ -83,17 +83,17 @@ func (t *collectDatabaseTask) validateInputs() error {
 // starting time in the Experiment
 func getElapsedTime(versionInfo map[string]interface{}, exp *Experiment) (int64, error) {
 	// ElapsedTime should not be provided by the user
-	if versionInfo[ElapsedTime] != nil {
+	if versionInfo[elapsedTimeStr] != nil {
 		return 0, errors.New("ElapsedTime should not be provided by the user in VersionInfo: " + fmt.Sprintf("%v", versionInfo))
 	}
 
 	startingTime := exp.Result.StartTime.Unix()
-	if versionInfo[StartingTime] != nil {
+	if versionInfo[startingTimeStr] != nil {
 		// Calling Parse() method with its parameters
-		temp, err := time.Parse(TimeLayout, fmt.Sprintf("%v", versionInfo[StartingTime]))
+		temp, err := time.Parse(timeLayout, fmt.Sprintf("%v", versionInfo[startingTimeStr]))
 
 		if err != nil {
-			return 0, errors.New("Cannot parse startingTime")
+			return 0, errors.New("cannot parse startingTime")
 		} else {
 			startingTime = temp.Unix()
 		}
@@ -210,7 +210,7 @@ func (t *collectDatabaseTask) run(exp *Experiment) error {
 			if err != nil {
 				return err
 			}
-			versionInfo[ElapsedTime] = elapsedTime
+			versionInfo[elapsedTimeStr] = elapsedTime
 
 			// finalize metrics template
 			template, err := template.ParseFiles(provider + ".metrics.yaml")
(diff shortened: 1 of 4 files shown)


base/collect_database.go
```diff
@@ -215,28 +219,6 @@ func (t *collectDatabaseTask) run(exp *Experiment) error {
...
-	}
-
-	// collect all files paths in current directory that ends with metrics.yaml
-	metricFilePaths := []string{}
-	err = filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
-		}
-
-		if !info.IsDir() && strings.HasSuffix(path, "metrics.yaml") {
```
Alan-Cha, line 231: The old method of identifying metrics files is to iterate through all file paths and look for paths that end with `metrics.yaml`.

base/collect_database.go
```diff
@@ -245,7 +227,7 @@ func (t *collectDatabaseTask) run(exp *Experiment) error {
 	}
 
 	// collect metrics for all metric files and versionInfos
-	for _, metricFilePath := range metricFilePaths {
+	for _, provider := range t.With.Providers {
```
Alan-Cha, line 206: The new method is to use the `providers`.

sriumcp:
Could we please change the struct field name `Url` to `URL`? The JSON and yaml tags (which are currently `url`) is perfect and can be retained.

base/collect_database.go
```diff
@@ -115,6 +114,7 @@ func getElapsedTime(versionInfo map[string]interface{}, exp *Experiment) (int64,
 		case int64:
 			startingTime = rawStartingTime
 		case int:
+		case float64: // parsing the metrics file gives float64
```
sriumcp: I think we want startingTime be of type string, and not any other type.
***
This way, it can be deserialized as follows ... 
```go

    // Declaring layout constant
    const layout = "Jan 2, 2006 at 3:04pm (MST)"
  
    // Calling Parse() method with its parameters
    startingTime, _ := time.Parse(layout, "Feb 4, 2014 at 6:05pm (PST)")
```

You can choose any layout of course, but it would probably best to begin with the default layout for time values in Golang. We can always adjust this part.
***

Wherever you compute elapsedTime, you can do:

```go
elapsedTime := startingTime.Sub(time.Now())
```

The above would make elapsedTime a [duration object](https://pkg.go.dev/time#Duration).

sriumcp (changes requested):
Please see comment on startingTime string type.

base/collect_database.go
kalantar: > I think we want startingTime be of type string, and not any other type.

Do we have control over this? The type for VersionInfo is `[]map[string]interface{}`. I agree that it would be better if the type were well defined.

base/collect_database.go
```diff
@@ -47,32 +45,14 @@ type Params struct {
...
 }
 
-const startingTimeString = "StartingTime"
-const elapsedTimeString = "ElapsedTime"
-
-// ToDo: Iter8 wiki is a great place to document the following (whichs seems IBM Code Engine specific) ... move it there
+const (
+	StartingTime = "StartingTime"
+	ElapsedTime  = "ElapsedTime"
+	TimeLayout   = "Jan 2, 2006 at 3:04pm (MST)"
```
sriumcp: Please change `TimeLayout` to `timeLayout`

sriumcp (changes requested):
Minor change requested.

base/collect_database.go
Alan-Cha: I can make this change. Is there a reason behind this though?

base/collect_database.go
```diff
@@ -47,32 +45,14 @@ type Params struct {
 	Value string `json:"value" yaml:"value"`
 }
 
-const startingTimeString = "StartingTime"
-const elapsedTimeString = "ElapsedTime"
-
-// ToDo: Iter8 wiki is a great place to document the following (whichs seems IBM Code Engine specific) ... move it there
+const (
+	StartingTime = "StartingTime"
```
sriumcp: `StartingTime` -- please change to `startingTime`.
`ElapsedTime` -- please change to `elapsedTime`.

***

In go, capitalizing a const or variable or func or type means you are exporting  (and exposing) it outside the package. In this case, there is no need, and these can be kept package local.

sriumcp (changes requested):

base/collect_database.go
```diff
@@ -45,31 +45,12 @@ type Params struct {
...
```
sriumcp: Perhaps:

```go
const (
  startingTimeStr = "StartingTime"
  elapsedTimeStr = "ElapsedTime"
  timeLayout = ... 
```

sriumcp (changes requested):

sriumcp (changes requested):
lgtm

sriumcp (approved):
lgtm

PR #1250, opened 2022-05-03
Document GitHub Action

kalantar:
Documents the GitHub action so that we can publish to marketplace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,19 @@
 
 Please see [https://iter8.tools](https://iter8.tools) for the complete documentation.
 
+## :white_check_mark: Installing Iter8 using GitHub Actions
+
+Install the latest version of Iter8 using the GitHub Action `iter8/iter8@v0.10`. A specific version can be installed using the version as the action reference. For example, to install version v0.10.15, use `iter8/iter8@v0.10.15`.
+
+Once Iter8 is installed, it can be used as documented (see [https://iter8.tools](https://iter8.tools)) in `run` actions. For example:
+
+```yaml
+- uses: iter8/iter8@v0.10 # install Iter8
+- run: |
+    iter8 version
+    iter8 launch -c load-test-http --set url=http://httpbin.org/get
+```
+
 ## :maple_leaf: Issues
 Iter8 issues are tracked [here](https://github.com/iter8-tools/iter8/issues).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,19 @@
 
 Please see [https://iter8.tools](https://iter8.tools) for the complete documentation.
 
+## :white_check_mark: Installing Iter8 using GitHub Actions
+
+Install the latest version of Iter8 using the GitHub Action `iter8/iter8@v0.10`. A specific version can be installed using the version as the action reference. For example, to install version v0.10.15, use `iter8/iter8@v0.10.15`.
```
Alan-Cha: I don't really like the following sentence:

```
A specific version can be installed using the version as the action reference
```

I didn't understand it at first and had to reread it multiple times. Maybe the following?

```suggestion
Install the latest version of Iter8 using the GitHub Action `iter8/iter8@v0.10`. A specific version can be installed using the version reference. For example, to install version v0.10.15, use `iter8/iter8@v0.10.15`.
```

sriumcp: > A specific version can be installed using the version as the action reference. For example, to install version v0.10.15, use `iter8/iter8@v0.10.15`.

I don't think we should say this at all. We don't want to tell users how to install patch versions in GitHub Actions, only how to install `major.minor` versions. The idea is that people will get the latest patch release (with bug fixes and security updates) without having to do any work themselves (which they would need to the moment they go for patch releases).

sriumcp (changes requested):
Please see comment on removing patch versioning.

README.md
sriumcp: Also, please see  `gh actions` tab [here](https://iter8.tools/0.10/getting-started/install/).

kalantar: per suggestion in [slack](https://iter8-tools.slack.com/archives/C01NV4T21HU/p1651589089239229?thread_ts=1651587863.205229&cid=C01NV4T21HU), moved this section to https://iter8.tools. See https://github.com/iter8-tools/docs/pull/29.  Simplified content in README.

sriumcp (changes requested):
Minor change: could please say "Use in GitHub Actions workflows"

sriumcp (approved):
lgtm

PR #1254, opened 2022-05-06
Istio SLO experiment gen and launch

Alan-Cha:


charts/iter8lib/templates/_metrics-istio.tpl
```diff
@@ -0,0 +1,206 @@
...
+#   source_cluster                            string
+#   source_principal                          string
+#   source_version                            string
+#   source_workload                           string
+#   source_workload_namespace                 string
+#
+# Inputs for the metrics (output of template):
+#   destination_workload                      string
+#   destination_workload_namespace            string
+#   StartingTime                              string
```
Alan-Cha, line 47: Should this also be lower case?

```suggestion
#   startingTime                              string
```

testdata/templates/ce.metrics.tpl
```diff
@@ -9,7 +9,7 @@ headers:
   # to get the GUID, run: ibmcloud resource service-instance <NAME> --output json | jq -r '.[].guid'
   # https://cloud.ibm.com/docs/monitoring?topic=monitoring-mon-curl
   IBMInstanceID: {{ .GUID }}
-provider: IBM Cloud Code Engine Sysdig
+provider: ce
```
Alan-Cha, line 12: @sriumcp Can you clarify the usage of this field? It should be `ce` to match the `providers` used in for example initializing the experiment, right?

charts/iter8lib/templates/_k-spec-secret.tpl
```diff
@@ -7,5 +8,13 @@ metadata:
     iter8.tools/revision: {{ .Release.Revision | quote }}
 stringData:
   experiment.yaml: |
-{{- include "experiment" . | indent 4 }}
+{{ include "experiment" . | indent 4 }}
+{{- if .Values.providers }}
+  metrics.yaml: |
+{{- range .Values.providers }}
+{{- if eq . "istio"}}
```
Alan-Cha, line 15: @sriumcp I was able to implement what you wanted, which was to dynamically construct the name of the template. 

This doesn't seem to be a common pattern. [This](https://stackoverflow.com/questions/20716726/call-other-templates-with-dynamic-name) was all I could find on the subject and it seems to be a frowned upon practice.

base/collect_database_test.go
```diff
@@ -10,13 +11,26 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var templatePath string = "../testdata/templates/ce.metrics.tpl"
-var tempMetricsPath string = "test-ce.metrics.yaml"
+const (
+	templatePath      = "../testdata/templates/ce.metrics.tpl"
```
Alan-Cha, line 15: @sriumcp I wanted to change the `templatePath` so that it points to the template in `charts/iter8lib/templates` rather than relying on this old one in `testdata/templates` but I encountered some issues. (`testdata/templates` only contains this template by the way. I was always planning on using it temporarily.)

Firstly, because my metrics template uses the `include` action, I think I have to use the Helm's chart parser (chartutil?) as well as the `loader` package. I didn't know if you wanted these extra dependencies in `base/`. Secondly, this forces the test to rely on the `charts/` folder, and I don't know if `charts/` folder will always be in the same location, how to get that location, and if its contents will be complete. Thoughts?

action/gen_test.go
```diff
@@ -45,3 +46,30 @@ func TestGenGRPC(t *testing.T) {
 	s := m["proto"].(string)
 	assert.Equal(t, "helloworld.proto", s)
 }
+
+func TestGenDB(t *testing.T) {
```
Alan-Cha, line 50: This is a test I added. I am not sure if this is sufficient. Please let me know your thoughts.

action/gen.go
```diff
@@ -15,7 +16,10 @@ import (
 	"helm.sh/helm/v3/pkg/getter"
 )
 
-const chartsFolderName = "charts"
+const (
```
sriumcp, line 19: Perhaps retain `chartsFolderName`

action/gen.go
```diff
@@ -39,7 +43,7 @@ func NewGenOpts() *GenOpts {
 
 // chartDir returns the path to chart directory
 func (gen *GenOpts) chartDir() string {
-	return path.Join(gen.ChartsParentDir, chartsFolderName, gen.ChartName)
+	return path.Join(gen.ChartsParentDir, charts, gen.ChartName)
```
sriumcp, line 46: Revert here also.

action/gen.go
```diff
@@ -94,5 +125,17 @@ func (gen *GenOpts) LocalRun() error {
 	}
 	log.Logger.Infof("created %v file", driver.ExperimentSpecPath)
 
+	// write metric spec files
+	for _, provider := range providers {
+		metricsBytes := []byte(m[path.Join(c.Name(), "templates", provider, driver.ExperimentMetricsPath)])
+		metricsFileName := provider + ".metrics.yaml"
```
sriumcp, line 131: Perhaps `const metricsYamlSuffix=".metrics.yaml"`?

base/collect_database_test.go
sriumcp: Use the `*.metrics.yaml` file for the unit test, and not the `*.metrics.tpl` file, since the unit tested functionality deals only with `yaml` and not `tpl`.

charts/iter8lib/templates/_k-spec-secret.tpl
sriumcp: We don't need to worry about `iter8lib` at all ... at the moment. Especially, changes to secret, cronjob, etc. can be handled in a separate PR.

charts/iter8lib/templates/_metrics-common-istio.tpl
```diff
@@ -0,0 +1,116 @@
+{{- define "metrics.common.istio" }}
```
sriumcp, line 1: We do need tpl files. Perhaps, create a `testdata/sample-chart/templates/istio.metrics.tpl` file for testing purposes.

charts/iter8lib/templates/_metrics-istio.tpl
```diff
@@ -0,0 +1,206 @@
+{{- define "metrics.istio" -}}
```
sriumcp, line 1: Perhaps create a chart under `testdata`.

charts/iter8lib/templates/_task-assess.tpl
```diff
@@ -10,5 +10,4 @@
       upperLimit: {{ $value | float64 }}
     {{- end }}
 {{- end }}
-{{- end }}
-
+{{- end }}
```
sriumcp, line 13: Perhaps leave all `iter8lib` changes out of this PR.

charts/load-test-http/values.yaml
```diff
@@ -10,7 +10,7 @@
 
 ### In addition, this chart supports the following parameters.
 
-### payloadUrl    URL of payload. If this field is specified, 
+### payloadURL    URL of payload. If this field is specified,
```
sriumcp, line 13: Documentation bug fix. Lets leave this in here.

charts/slo-validation-istio/.helmignore
```diff
@@ -0,0 +1,26 @@
+# Patterns to ignore when building packages.
```
sriumcp, line 1: This whole chart moves to `testdata`.

driver/common.go
```diff
@@ -9,6 +9,8 @@ import (
 const (
 	// ExperimentSpecPath is the name of the experiment spec file
 	ExperimentSpecPath = "experiment.yaml"
+	// ExperimentMetricsPath is the name of the metrics spec file
```
sriumcp, line 12: `ExperimentMetricsPathSuffix`

testdata/templates/ce.metrics.tpl
sriumcp: Yes. This is a descriptive tag for end users. Of course, we want to match providers in the chart values with this provider name.

sriumcp (changes requested):

Alan-Cha:
See #1255

PR #1255, opened 2022-05-06
SLO validation gen and launch

Alan-Cha:


base/collect_database_test.go
```diff
@@ -33,32 +29,6 @@ type collectDatabaseTemplateInput struct {
 	GUID     string `json:"GUID" yaml:"GUID"`
 }
 
-// has to be a map[string]string in order to do input checks in template
-// func executeTemplate(inputs interface{}, templatePath string, writePath string) error {
-func executeTemplate(inputs map[string]interface{}, templatePath string, writePath string) error {
```
Alan-Cha, line 38: Unfortunately, by simplifying the DB task test (no longer generates the metrics file from a template), I have to put the metrics file in `base/` because of how the DB task works. It looks for the metrics file in the current directory and in the context of the test, it is in `base/`. Thoughts?

Alan-Cha:
Also, by moving the db task and metrics template files to `testdata/`, I had to replicate the majority of `iter8lib/` in `testdata/` as well in order to make the gen test to work. Thoughts?

Feel free to try:

```
cd testdata

iter8 gen -c slo-validation-istio --set endpoint=http://127.0.0.1:9090 --set destination_workload=httpbin-v1 --set destination_workload_namespace=default --set StartingTime="Feb 4\, 2014 at 6\:05pm \(PST\)"
```

testdata/charts/slo-validation-istio/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+source_version: null
+source_workload: null
+source_workload_namespace: null
+
+##################################
+# Used by metrics
+
+destination_workload: null
+destination_workload_namespace: null
+StartingTime: null
```
Alan-Cha: In my previous PR, I asked about the capitalization of `StartingTime`. I don't think we addressed it this morning. Thoughts?

base/collect_database_test.go
Alan-Cha: `base/test-ce.metrics.yaml`

testdata/charts/slo-validation-istio/values.yaml
sriumcp: I think `startingTime` is the usual convention for Helm values. So, perhaps camelCase?

sriumcp:
`iter8lib` in `testdata` is totally fine for now. We can cleanup in the next PR. Whatever is needed to enable tests for now can go into `testdata` of course.

base/collect_database_test.go
sriumcp: I feel we can cleanup the tests a bit to address this in the short term, and cleanup the file driver and kubedriver to address this in the next PR.

base/collect_database.go
```diff
@@ -57,8 +57,10 @@ type collectDatabaseInputs struct {
 }
 
 const (
-	// collectDatabaseTaskName is the name of this task which performs load generation and metrics collection for gRPC services.
+	// CollectDatabaseTaskName is the name of this task which performs load generation and metrics collection for gRPC services.
 	CollectDatabaseTaskName = "collect-metrics-database"
+	// experimentMetricsPathSuffix is the name of the metrics spec file
+	experimentMetricsPathSuffix = ".metrics.yaml"
```
sriumcp, line 63: I see two constants ... "experimentMetricsPathSuffix" and "ExperimentMetricsPathSuffix"... combine?

base/collect_database_test.go
sriumcp: Perhaps, we should put this file under `testdata/metrics` folder, do a `os.Chdir(t.TempDir())` within the unit tests, and then copy the `base/test-ce.metrics.yaml` file into the tempDir?

```go
// approximate code
destDir := t.TempDir()
os.Chdir(destDir)
src := base.CompletePath("../testdata", "metrics/test-ce.metrics.yaml")
io.Copy(src, base.CompletePath(destDir, metricYamlFileName))
// rest of the test goes here... 
```

sriumcp (changes requested):

testdata/charts/slo-validation-istio/values.yaml
Alan-Cha: Ok

base/test-ce.metrics.yaml
```diff
@@ -0,0 +1,77 @@
...
+url: test-database.com/prometheus/api/v1/query # e.g. https://ca-tor.monitoring.cloud.ibm.com
+headers:
+  # IAM token
+  # to get the token, run: ibmcloud iam oauth-tokens | grep IAM | cut -d \: -f 2 | sed 's/^ *//'
+  Authorization: Bearer test-token
+  # GUID of the IBM Cloud Monitoring instance
+  # to get the GUID, run: ibmcloud resource service-instance <NAME> --output json | jq -r '.[].guid'
+  # https://cloud.ibm.com/docs/monitoring?topic=monitoring-mon-curl
+  IBMInstanceID: test-guid
+provider: IBM Cloud Code Engine Sysdig
```
Alan-Cha: This needs to be changed to `ce`

base/collect_database.go
Alan-Cha: @sriumcp I forgot to point this out. If I were to use `ExperimentMetricsPathSuffix` here, I would need to bring in `driver` and it would give me an import cycle error. I was not sure what the best way to resolve this error is so I went with this solution for now.

base/collect_database.go
```diff
@@ -46,7 +46,7 @@ type Params struct {
 }
 
 const (
-	startingTimeStr = "StartingTime"
+	startingTimeStr = "startingTime"
 	elapsedTimeStr  = "ElapsedTime"
```
Alan-Cha: @sriumcp Should `ElapsedTime` also be renamed to `elapsedTime` as well?

sriumcp: That seems like a good idea ...

base/collect_database.go
```diff
@@ -99,7 +101,7 @@ func getElapsedTime(versionInfo map[string]interface{}, exp *Experiment) (int64,
 		}
 	}
 
-	// calculate the ElapsedTime based on the StartingTime if it has been provided
+	// calculate the ElapsedTime based on the startingTime if it has been provided
```
sriumcp: What is the units for `elapsedTime`? Is it seconds? Nanoseconds? Something else?

base/collect_database.go
sriumcp: We will resolve this in a subsequent PR with changes too drivers.

sriumcp (changes requested):

base/collect_database.go
kalantar: agree this should be documented

sriumcp (approved):
lgtm

PR #1260, opened 2022-05-17
Add metrics spec to driver

Alan-Cha:
Signed-off-by: Alan Cha <Alan.cha1@ibm.com>

### DIFF
--- a/driver/common.go
+++ b/driver/common.go
@@ -1,6 +1,8 @@
 package driver
 
 import (
+	"text/template"
+
 	"github.com/iter8-tools/iter8/base"
 	"github.com/iter8-tools/iter8/base/log"
 	"sigs.k8s.io/yaml"
@@ -26,6 +28,11 @@ func SpecFromBytes(b []byte) (base.ExperimentSpec, error) {
 		return nil, err
 	}
 	return e, err
+}
+
+// MetricsSpecFromBytes reads metrics spec from bytes
+func MetricsSpecFromBytes(b []byte) (*template.Template, error) {
+	return template.New("metrics-spec").Parse(string(b))
 }
 
 // ResultFromBytes reads experiment result from bytes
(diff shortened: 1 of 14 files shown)


driver/common.go
```diff
@@ -28,6 +30,11 @@ func SpecFromBytes(b []byte) (base.ExperimentSpec, error) {
 	return e, err
 }
 
+// MetricsSpecFromBytes reads metrics spec from bytes
+func MetricsSpecFromBytes(b []byte) (*template.Template, error) {
+	return template.Must(template.New("metrics-spec").Parse(string(b))), nil
```
sriumcp: Wrapping the `(*template, err)` using `template.Must` will end up converting error into a panic!

I think it would be better to simply return `template.New("metrics-spec").Parse(string(b))` (without the `template.Must` wrapping).

driver/kubedriver.go
```diff
@@ -161,6 +162,12 @@ func (driver *KubeDriver) getSpecSecretName() string {
 	return fmt.Sprintf("%v-spec", driver.Group)
 }
 
+// getMetricsSpecSecretName yields the name of the experiment metrics spec
+// secret
+func (driver *KubeDriver) getMetricsSpecSecretName(provider string) string {
```
sriumcp: I don't think we want a separate secret for metrics. 

We will use the same secret used for storing the experiment spec. This will give us two secrets always, one for experiment/metric specs, another for result.

sriumcp (changes requested):

sriumcp:
@Alan-Cha The unit test is failing for unrelated reasons as you pointed out.

In this regard... I have a request.

Could you please remove [lines 17 and 18 -- i.e., the 2nd responder](https://github.com/iter8-tools/iter8/blob/34b2cfe793fd185ddfa6a2391d1fe1a742d88a7f/base/collect_http_test.go#L17) from the test file whenever you update the PR? I think that might fix the issue.

sriumcp (approved):
lgtm

kalantar (approved):
lgtm

PR #1270, opened 2022-06-06
Add cronjob template

Alan-Cha:


.github/workflows/tests.yaml
```diff
@@ -212,6 +211,7 @@ jobs:
       run: |
         iter8 k launch -c load-test-http --noDownload \
         --set url=http://httpbin.default \
+        --set iter8lib.runner=job \
```
sriumcp: Could you please update the `load-test-http` and `load-test-grpc` values.yaml files so that we have:
```yaml
iter8lib:
  runner: job
```

in their respective values file? 

***

This way, in the second test (and in the documentation), there would be no need for the user to set this somewhat arcane value.

But in the first test, you can explicitly null it `--set iter8lib.runner=none`.

***

In the `istio-slo-validation` values.yaml file, you will of course have:
```yaml
iter8lib:
  runner: cronjob
```

charts/iter8lib/templates/_k-cronjob.tpl
```diff
@@ -0,0 +1,25 @@
+{{- define "k.cronjob" -}}
```
sriumcp, line 1: Could you please see concurrency policy to "forbid"?

Please see details [here](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#concurrency-policy).

sriumcp (changes requested):
Two minor changes requested.

charts/iter8lib/values.yaml
```diff
@@ -4,11 +4,15 @@ iter8Image: iter8/iter8:0.10
...
+### disable manifests that should be disabled. This is useful for testing purposes.
 ### By default, no manifest is disabled.
-### To disable the job manifest, uncomment the next two lines and delete the line after that.
-### disable:
-###   job: true
 ### To disable rbac role and rolebinding for readiness, uncomment the next line
 ###   readinessrbac: true
 disable: {}
+
+### enable to use job or a cronjob
```
Alan-Cha, line 13: Is this comment sufficient? I wasn't exactly sure how to describe it.

kalantar (commented):
looks good to me.

charts/iter8lib/values.yaml
sriumcp: lgtm for now.

sriumcp (approved):
lgtm

PR #1278, opened 2022-06-09
Assorted changes for SLO validation cronjob

Alan-Cha:


base/experiment.go
```diff
@@ -823,12 +833,13 @@ func BuildExperiment(withResult bool, driver Driver) (*Experiment, error) {
...
 		}
 	}
+	e.numLoops = 1
 	return &e, nil
 }
 
 // RunExperiment runs an experiment
-func RunExperiment(driver Driver) error {
-	if exp, err := BuildExperiment(false, driver); err != nil {
+func RunExperiment(driver Driver, reuseResult bool) error {
```
Alan-Cha: Should I reverse the order of these arguments to match `BuildExperiment()`? Also what about the name of the argument? `reuseResult` vs `withResult` (like in in `BuildExperiment()`)? I feel like the former makes more sense in this context.

cmd/run.go
```diff
@@ -32,6 +32,7 @@ func newRunCmd(kd *driver.KubeDriver, out io.Writer) *cobra.Command {
 		},
 	}
 	addRunDirFlag(cmd, &actor.RunDir)
+	addReuseResult(cmd, &actor.ReuseResult)
```
Alan-Cha, line 35: I don't remember if we discussed to add the `reuseResult` flag to both `iter8 k run` and `iter8 run`. I think maybe it should only be for `iter8 k run`? Then I will need to change this.

charts/iter8lib/templates/_k-cronjob.tpl
```diff
@@ -20,7 +23,7 @@ spec:
             - "/bin/sh"
             - "-c"
             - |
-              iter8 k run --namespace {{ .Release.Namespace }} --group {{ .Release.Name }}
+              iter8 k run --namespace {{ .Release.Namespace }} --group {{ .Release.Name }} -l {{ .Values.iter8lib.logLevel }} --reuseResult
```
Alan-Cha: Maybe I should also change the order so that log level is at the end.

cmd/run.go
```diff
@@ -40,6 +41,12 @@ func addRunDirFlag(cmd *cobra.Command, runDirPtr *string) {
 	cmd.Flags().StringVar(runDirPtr, "runDir", ".", "directory where experiment is run; contains experiment.yaml and result.yaml")
 }
 
+// addReuseResult allows the experiment to reuse the experiment result for
+// looping experiments
+func addReuseResult(cmd *cobra.Command, reuseResultPtr *bool) {
+	cmd.Flags().BoolVar(reuseResultPtr, "reuseResult", false, "reuse experiment result for looping experiment")
```
Alan-Cha, line 47: I wasn't sure if we wanted a shorthand option. We currently don't have many other options for `iter8 k run` but I feel that this might be too niche to warrant it.

cmd/run.go
sriumcp: Good for now.

cmd/run.go
sriumcp: agreed

charts/iter8lib/templates/_k-cronjob.tpl
sriumcp: sounds good

sriumcp (approved):
lgtm

PR #1284, opened 2022-06-14
Iter8 experiment chart

### LINKED ISSUE #1272: Iter8 report should not have histogram section when there are no histograms
report html

### LINKED ISSUE #1279: `iter8 k run` `group` flag should have default but does not work
```bash
Alans-MBP:iter8-alan alan.cha$ iter8 k run -h

Run a Kubernetes experiment. This command reads an experiment specified in a secret and writes the result to another secret.

	$ iter8 k run --namespace {{ .Experiment.Namespace }} --group {{ .Experiment.group }}

This command is intended for use within the Iter8 Docker image that is used to execute Kubernetes experiments.

Usage:
  iter8 k run [flags]

Flags:
  -g, --group string   name of the experiment group (default "default")
  -h, --help           help for run
      --reuseResult    reuse experiment result for looping experiment

Global Flags:
      --kube-apiserver string       the address and the port for the Kubernetes API server
      --kube-as-group stringArray   group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string         username to impersonate for the operation
      --kube-ca-file string         the certificate authority file for the Kubernetes API server connection
      --kube-context string         name of the kubeconfig context to use
      --kube-token string           bearer token used for authentication
      --kubeconfig string           path to the kubeconfig file
  -l, --loglevel string             trace, debug, info, warning, error, fatal, panic (default "info")
  -n, --namespace string            namespace scope for this request
Alans-MBP:iter8-alan alan.cha$ iter8 k run --reuseResult
Error: required flag(s) "group" not set
```

Ignore the `reuseResult` flag but note how the documentation says the default value is `default` but I get an error that says `group` is not set.

### LINKED ISSUE #1281: Simplify experiment chart
We have three experiment charts at the moment ... all of which are very thin wrappers around `iter8lib`. This feels like an overkill, especially, when it comes to documenting how to extend Iter8. For example, in order to extend the Istio SLO validation to Linkerd, a user should not have to write chart, but should be able to get away by creating a metrics template file, instead of creating a whole new chart.

Create a single `iter8` chart that works for all experiments.

For example:

```shell
iter8 k launch \
--set tasks={custommetrics,assess} \
--set custommetrics.providersURLs={"https://github.com/myorg/myproj/istio.tpl","https://github.com/yourorg/yourproj/path/to/kserve.tpl"} \
--set assess.SLOs.istio/mean-latency=50 \
--set assess.SLOs.kserve/accuracy=0.9 \
--set cronschedule="some cron schedule"
```

```shell
iter8 k launch \
--set tasks={ready,http,assess} \
--set ready.deploy=httpbin \
--set ready.service = httpbin \
--set http.url=http://hello.default/get \
--set assess.SLOs.http/latency-mean=50 \
--set assess.SLOs.http/latency-p50=50
```

### LINKED ISSUE #1283: Metrics template URL
If a user needs to use SLO validation features with custom metrics, currently they have to touch iter8 experiment charts. This is non-trivial and complicated from the perspective of a developer.

Change the `read-metrics-from-db` task to be able to read provider templates from URLs. This way, the only thing a template author needs to do is to upload the provider template onto a URL, and use it as part of the launch command.

For example, 

```shell
iter8 k launch \
--set tasks={custommetrics,assess} \
--set custommetrics.providersURLs={"https://github.com/myorg/myproj/istio.tpl","https://github.com/yourorg/yourproj/path/to/kserve.tpl"} \
--set assess.SLOs.istio/mean-latency=50 \
--set assess.SLOs.kserve/accuracy=0.9 \
--set cronschedule="some cron schedule"
```

### LINKED ISSUE #1288: `gh-action` test needs to happen post release
`gh-action` depends on release assets, which are created after a release event. So, it makes sense to do `gh-action` tests after releases. If it is done before releases, the test passes for non-breaking changes, but fails (understandably) for breaking changes.

### LINKED ISSUE #1289: Standardize parameters for tasks and make task names short
There is no way to tell which parameter is for which task, which complicates task templates; this also makes the `iter8 k launch` commands a bit harder to parse (explicitly stating the task name as part of the parameter setting makes both easier).

Example below...

```shell
iter8 launch --noDownload \
         --set "tasks={http,assess}" \
         --set http.url="http://$HOST_IP/post" \
         --set http.payloadURL="https://httpbin.org/stream/1" \
         --set http.duration="3s" \
         --set assess.SLOs.upper.http/latency-p95=500 \
         --set assess.SLOs.upper.http/error-count=0
```

We sometime dump `.Values` into the `with` clause of tasks within templates. This makes the task writing simpler, but the resultant experiment.yaml is bloated needlessly. The task template prefixes for values gives us the best of both worlds.

### LINKED ISSUE #1290: Collapse experiment spec and experiment result into a single artifact
Two artifacts, one for spec and one for result, means replication of code (for roles; secrets; role bindings; reading and writing).

Collapse them into a single artifact. This does mean ... the artifact under question (`experiment.yaml`) could change during an experiment after it is initially specified. This is no longer a problem because the user is not intended to directly deal with `experiment.yaml`.

sriumcp:
Fixes #1281 
Fixes #1283
Fixes #1279 
Fixes #1272 
Fixes #1288
Fixes #1289
Fixes #1290


.github/workflows/assets.yaml
```diff
@@ -93,12 +93,46 @@ jobs:
         build-args: |
           TAG=v${{ env.VERSION }}
 
+  gh-action:
```
sriumcp, line 96: gh-action test needs to run after the release assets are created. Hence moving them.

.github/workflows/assets.yaml
```diff
@@ -110,8 +144,10 @@ jobs:
         kubectl wait --for=condition=available --timeout=60s deploy/httpbin
     - name: iter8 k launch
       run: |
-        iter8 k launch -c load-test-http \
-        --set url="http://httpbin.default/get"
+        iter8 k launch \
+        --set tasks={http} \
```
sriumcp, line 148: Notice how `--set tasks={http}` replaces the `--chartName load-test-http` mode of creating experiments.
***
There is only one experiment chart now, helpfully named `iter8`, and no longer required to be specified by the user. You can flexibly compose an experiment by simply specifying the tasks... 

For e.g., `--set tasks={http,assess,notify}` would put together an experiment with these three tasks (`notify` is yet-to-be created).

.github/workflows/assets.yaml
```diff
@@ -141,13 +174,45 @@ jobs:
...
         iter8 k report
         iter8 k log
         iter8 k delete
+
+  readiness:
+    name: Kubernetes readiness test
+    runs-on: ubuntu-latest
+    needs: 
+    - build-and-push
+    - assets
```
sriumcp, line 195: Needs assets.

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update && apt-get install -y wget
 
 # Set Iter8 version from build args
 ARG TAG
-ENV TAG=${TAG:-v0.10.15}
+ENV TAG=${TAG:-v0.11.0}
```
sriumcp, line 9: breaking changes

action/assert.go
```diff
@@ -75,7 +75,7 @@ func (assert *AssertOpts) verify(eio base.Driver) (bool, error) {
 
 	// check assert conditions
 	for {
-		exp, err := base.BuildExperiment(true, eio)
+		exp, err := base.BuildExperiment(eio)
```
sriumcp, line 78: Experiment build will always read results. But they may or may not be reused during the run (based on the --reuseResult flag). For looped cron-job experiments, `reuseResult` is true.

action/assert_test.go
```diff
@@ -15,9 +16,10 @@ import (
 )
 
 func TestLocalAssert(t *testing.T) {
+	os.Chdir(t.TempDir())
```
sriumcp, line 19: It is safe to change to a tempdir always in every single test. This is because, tests might involve creating/manipulating files like `experiment.yaml`. We don't want to mess up in `testdir` folders accidentally.

action/delete_test.go
```diff
@@ -15,9 +15,9 @@ func TestKubeDelete(t *testing.T) {
 	// fix lOpts
 	lOpts := NewLaunchOpts(driver.NewFakeKubeDriver(cli.New()))
 	lOpts.ChartsParentDir = base.CompletePath("../", "")
-	lOpts.ChartName = "load-test-http"
+	lOpts.ChartName = "iter8"
 	lOpts.NoDownload = true
-	lOpts.Values = []string{"url=https://iter8.tools", "duration=2s"}
+	lOpts.Values = []string{"tasks={http}", "http.url=https://iter8.tools", "http.duration=2s"}
```
sriumcp, line 20: `http` task template values will be under `.Values.http.`; inspiration is from `ready` task template, whose values are all under the `.Values.ready.`. Same is true for all task templates (assess, grpc, etc.)

action/gen.go
```diff
@@ -18,7 +17,7 @@ import (
 
 const (
 	chartsFolderName = "charts"
-	providersStr     = "providers"
+	DefaultChartName = "iter8"
```
sriumcp, line 20: Users no longer need to provide chart names.

action/gen.go
```diff
@@ -63,37 +63,10 @@ func (gen *GenOpts) LocalRun() error {
 	// add in experiment.yaml template
 	eData := []byte(`{{- include "experiment" . }}`)
 	c.Templates = append(c.Templates, &chart.File{
-		Name: path.Join("templates", driver.ExperimentSpecPath),
+		Name: path.Join("templates", driver.ExperimentPath),
```
sriumcp, line 66: There are no spec and result files/secrets anymore. One object handles both.

action/report/htmlreport.tpl
```diff
@@ -78,27 +78,42 @@
               </tr>
             </thead>
             <tbody>
-                {{- range $ind, $slo := .Result.Insights.SLOs }}
+                {{- range $ind, $slo := .Result.Insights.SLOs.Upper }}
```
sriumcp, line 81: This change enables users to specify both upper and lower limits for SLOs very easily. Previous template only supported upper.

base/collect_grpc.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const (
 	// CollectGRPCTaskName is the name of this task which performs load generation and metrics collection for gRPC services.
-	CollectGRPCTaskName = "gen-load-and-collect-metrics-grpc"
+	CollectGRPCTaskName = "grpc"
```
sriumcp, line 14: Short names are easier on the documentation. Task template `_task-grpc`, task name `grpc`, chart values `.Values.grpc.call`, etc., can also be made concise and consistent with short names.

base/custom_metrics.go
```diff
@@ -71,42 +72,45 @@ type Params struct {
 	Value string `json:"value" yaml:"value"`
 }
 
-// collectDatabaseInputs is the input to the collect-metrics-database task
-type collectDatabaseInputs struct {
-	// Providers is the set of labels/names of the data sources
-	Providers []string `json:"providers" yaml:"providers"`
+// customMetricsInputs is the input to the custommetrics task
```
sriumcp, line 75: We want to call this task `custommetrics` which is what it really is (as opposed to built-in metrics).

base/custom_metrics.go
```diff
@@ -71,42 +72,45 @@ type Params struct {
...
+type customMetricsInputs struct {
+	// ProviderURLs a slice of URLs for metric templates
+	ProviderURLs []string `json:"providerURLs" yaml:"providerURLs"`
```
sriumcp, line 78: Provider templates are now hosted in URLs instead of supplied as files (within charts). This means any user can simply put their template in any URL (e.g. their GH repo) and start using it in Iter8 experiments.

base/experiment.go
```diff
@@ -34,19 +33,22 @@ type ExperimentSpec []Task
...
-	Result *ExperimentResult
+	Result *ExperimentResult `json:"result" yaml:"result"`
 
 	// driver enables interacting with experiment result stored externally
 	driver Driver
 }
 
 // ExperimentResult defines the current results from the experiment
 type ExperimentResult struct {
+	// Revision of this experiment
```
sriumcp, line 49: Idea is to store experiment revision in the experiment result. This is still not fully tested... will be tested with autox.

charts/iter8/templates/_experiment.tpl
```diff
@@ -0,0 +1,26 @@
+{{- define "experiment" -}}
```
sriumcp, line 1: This enables the user to compose a new experiment type without having to create (or even understand about ) experiment charts.

charts/iter8/templates/_k-job.tpl
```diff
@@ -13,15 +14,16 @@ spec:
       annotations:
         sidecar.istio.io/inject: "false"
     spec:
+      serviceAccountName: {{ .Release.Name }}-iter8-sa
```
sriumcp, line 17: Always run with a proper service account.

cmd/k.go
```diff
@@ -12,12 +12,9 @@ var kCmd = &cobra.Command{
 	Long:  "Work with Kubernetes experiments",
 }
 
-// addExperimentGroupFlag adds the experiment group flag and marks it as required or optional
-func addExperimentGroupFlag(cmd *cobra.Command, groupP *string, required bool) {
+// addExperimentGroupFlag adds the experiment group flag
+func addExperimentGroupFlag(cmd *cobra.Command, groupP *string) {
```
sriumcp, line 16: Group flag becomes optional

custommetrictemplates/istio.tpl
```diff
@@ -1,5 +1,4 @@
-{{- define "metrics.istio" -}}
-url: {{ .Values.providerURL }}
+url: {{ .providerURL }}
```
sriumcp, line 1: This metrics template will be supplied to experiments as a URL value.

kalantar (dismissed):
thanks for the walkthrough

kalantar (approved):

PR #1304, opened 2022-06-29
Custom metrics fixes

### LINKED ISSUE #1300: Prevent overwrite of the `charts` folder
Iter8 commands like `launch`, and `k launch` have a need to have the following behavior for the `charts` folder.

If `charts` folder is already exists inside the `chartsParentDir`, then the `--noDownload` flag should become mandatory (i.e., it should be set to true or false). 

This will help avoid subtle issues like #1299

### LINKED ISSUE #1302: Change `custom metrics` task inputs to be more user friendly
Specifying multiple templates for custom metrics looks as follows today:

```shell
--set custommetrics.providerURLs[0]=..
--set custommetrics.providerURLs[1]=..
```

If two of these templates have the same provider names, the task might fail or will need to resolve the naming conflict in some round about manner.

Adopt the following task signature.

```shell
--set custommetrics.templates.istio-prom=<istio-prom-template-url>
--set custommetrics.templates.knative-prom=<knative-prom-template-url>
```

First of all, we renamed the field as `templates`, which is explicit (it is about templates, and does not refer to the URL of the provider database -- only the provider template). Second, my making it a map, we circumvent naming conflicts (a map will have unique keys by definition). Third, map values are generally encouraged over array values by Helm.

sriumcp:
Fixes #1302 
Fixes #1300


kalantar (dismissed):
lgtm

kalantar (approved):

PR #1313, opened 2022-07-29
Iter8 A/B(/n) service

kalantar:


abn/metricstore/summary_metric.go
```diff
@@ -0,0 +1,95 @@
...
+	(*m)[SS_IDX] = v
+}
+
+func (m *SummaryMetric) LastUpdateTimestamp() time.Time {
+	return time.Unix(int64(math.Round((*m)[TS_IDX])), 0)
+}
+
+func (m *SummaryMetric) SetLastUpdateTimestamp(t time.Time) {
+	(*m)[TS_IDX] = float64(t.Unix())
+}
```
Alan-Cha: Why is it necessary to cast the Unix time stamp from `float64` back into `int64`?

abn/service.go
```diff
@@ -0,0 +1,141 @@
...
+
+const (
+	// Name of environment variable with file path to resources/namespaces to watch
+	WATCHER_CONFIG_ENV = "WATCHER_CONFIG"
+)
+
+var (
+	// Port the service listens on
+	port = flag.Int("port", 50051, "The server port")
+	// kubedriver *driver.KubeDriver = nil
```
Alan-Cha, line 33: In case you forgot about this

abn/util/util.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+const (
+	code = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+)
+
+// RandomString returns a random string of length characters
+func RandomString(length int) string {
+	ran_str := make([]byte, length)
+	// Generating Random string
```
Alan-Cha:
```suggestion
	// generate random string
```

abn/watcher/app.go
```diff
@@ -0,0 +1,192 @@
...
+	// Ready indicates if version is ready
+	Ready bool
+	// Track is track of version
+	Track string
+}
+
+// Apps is map of app name to Application
+var Apps map[string]Application = map[string]Application{}
+
+// recrodEvent ensures we call attempt to write events only when the recorder is non-null
```
Alan-Cha:
```suggestion
// recordEvent ensures we call attempt to write events only when the recorder is non-null
```

abn/watcher/config.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"io/ioutil"
+
+	"github.com/iter8-tools/iter8/base/log"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"sigs.k8s.io/yaml"
+)
+
+// Config is ABn serivce configuration
```
Alan-Cha:
```suggestion
// Config is A/B(/n) service configuration
```

abn/util/util.go
kalantar: done

abn/watcher/app.go
kalantar: done

abn/watcher/config.go
kalantar: done

abn/metricstore/summary_metric.go
kalantar: `math.Round` returns a `float64`; `time.Unix()` wants an `int64`. Hence the conversion.

Alan-Cha (dismissed):

abn/grpc/abn.proto
```diff
@@ -0,0 +1,51 @@
...
+
+option go_package = "github.com/iter8-tools/iter8/abn/grpc";
+
+import "google/protobuf/empty.proto";
+package main;
+
+// for more information, see https://github.com/iter8-tools/iter8/issues/1257
+
+service ABN {
+  // Identify a version the caller should send a request to.
```
sriumcp: Didn't we say that lookup returns a track and not a version?

abn/grpc/abn.proto
```diff
@@ -0,0 +1,51 @@
...
+  // name of (backend) application or service
+  // This value is used to identify the Kubernetes objects that make up the service
+  // Kubernetes objects that comprise the service should have the label app.kubernetes.io/name set to name
+  string name = 1;
+  // User or user session identifier
+  string user = 2;
+}
+
+message Session {
+  // Track or logical name of the application version
```
sriumcp: Could we please keep the tracks and versions as distinct ... and not meld them?

Track identifiers are static, and may be `baseline`, `candidate`, or `candidateX` (the first two are useful in the case of AB testing, and the first and the last track with X=0,1,2,... is useful in the case of ABN testing).

Version identifiers are free form -- the user can use git commit, cloud object IDs, image tags, ... however the user chooses to identify unique versions of an app/ML model.

abn/metricstore/metric.go
```diff
@@ -0,0 +1,258 @@
+package metricstore
```
sriumcp: This file is not strictly about metrics. It is also about the entire app history. So, perhaps instead of the folder being `metricsstore`, it can be `apphistory` or `appdata`.

abn/metricstore/metric.go
```diff
@@ -0,0 +1,258 @@
...
+
+type ApplicationData map[string]VersionData
+
+type VersionData struct {
+	Metrics             map[string]SummaryMetric `json:"metrics,omitempty" yaml:"metrics,omitempty"`
+	LastUpdateTimestamp time.Time                `json:"lastUpdateTimestamp" yaml:"lastUpdateTimestamp"`
+	History             []VersionEvent           `json:"history,omitempty" yaml:"history,omitempty"`
+}
+
+type VersionEvent struct {
```
sriumcp: I don't think we need to worry about Kubernetes objects that are improperly labeled/annotated. Iter8 should clearly document the labels/annotations it expects -- and if they are not present, the kubernetes object should be considered to be outside the purview of abn experimentation / service -- to be simply ignored by Iter8!
***
We are asking for a few labels, track, version, app, ready or not. This is really reasonable for a feature that deals with AB testing, and handles both metrics and routing.
***
We will use this approach for autox also -- if you supply the labels we require, some work will happen. Otherwise an expensive no-op will happen!

abn/metricstore/metric.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return &MetricStoreSecret{
+		App:       app,
+		Name:      nm,
+		Namespace: ns,
+		client:    kd.Clientset,
+	}, nil
+}
+
+type MetricStoreSecretCache struct {
```
sriumcp: It would be more intuitive, if we get rid of the cache struct altogether ... and define its methods directly on the ApplicationData struct... 

This way, you have methods to read and write ApplicationData object into the secret, and you also have methods to manipulate the ApplicationData object itself.

abn/service.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Lookup identifies a version that should be used for a given user
+// This method is exposed to gRPC clients
+func (server *abnServer) Lookup(ctx context.Context, a *pb.Application) (*pb.Session, error) {
+	v, err := pb.Lookup(a.GetName(), a.GetUser())
+	if err != nil {
+		return nil, err
+	}
+	track := v.Track
+	if track == "" {
```
sriumcp: Log an error instead. Please see discussion above on how tracks and versions are different concepts, and not the same.

sriumcp (changes requested):
Initial changes requested.

abn/grpc/abn.proto
kalantar: I have found it hard to define in a clear an understandable way the term "track'. I finally settled on the language "logical version". Is there a better way to define and describe this term?

kalantar: I did not understand that we wanted to restrict track names to a fixed static set. I did not implement it this way. Instead, the user may use any logical names they feel make sense to them. I can look at restricting the set of valid names.

abn/metricstore/metric.go
kalantar: agree. this name is bad. i will change the name but in my refactor for tasks, I eliminated this folder altogether.

abn/metricstore/metric.go
kalantar: Is this comment associated with the referenced code in some way? I do not see the relationship of the comment to the code. in general, this was the approach I took though as you do comment in a couple of other places, I did try to be a little flexible. Some of this was meant to be removed so ... thanks for catching.

abn/service.go
kalantar: agree.

abn/grpc/abn.proto
sriumcp: We want to fix track names, so that in the frontend code, the user can create a fixed track-to-route map. The frontend should not be affected when new versions of the backend are released. We achieve this by fixing the track -- and mapping tracks to routes (in the frontend, statically), and also mapping tracks to versions (in the abn service, dynamically, so that metrics can be associated with versions and not just tracks).

This is the main distinction between tracks and versions.

abn/metricstore/metric.go
sriumcp: I saw code that tried to assign version identifiers to tracks (in case the latter was missing). This is not desirable, since versions and tracks are distinct concepts. The right thing to do, IMO, if Kubernetes objects are not annotated properly, is to simply leave them alone (we can log ... eventually we can use a command to surface the abn logs easily to the user).

abn/grpc/abn.proto
sriumcp: **Definition of a track:** A track identifier is something which can be **statically** mapped to a route to a backend service. Multiple tracks and hence multiple routes may be available at any time. The route can be an HTTP route (URL with headers and auth values) or a gRPC route (gRPC call implemented service) -- entirely up to the app (in particular, the frontend) to decide what a route means.

The static mapping is critical in order to decouple the frontend code (the maps the result of the abnservice lookup call to a route) from backend version releases.

abn/metricstore/metric.go
kalantar: The primary motivator for this object was to cache the secret so that between the read and increment it could be written w/o rereading and parsing it again. On inspection, it seems pretty complicated. I agree that some simplification and clarification is a good idea.

sriumcp: To achieve this caching, could we have an in-memory map as follows:

```go
type AppDataMap map[string]ApplicationData
var appDataMap *AppDataMap
```

`appDataMap` is initialized from secrets (if any) when the abn service initializes.

```go
appDataMap = initFromSecrets()
```

 Whenever there is a need to read, the read directly happens from `ApplicationData`. Below is an example (which can of course be suitably modified if need be)

```go
appDataMap[appName].getVersionForTrack(trackName) // this method does not involve read from secrets; so yes, caching.
```

Whenever there is a need to write, you can directly update the `AppDataMap`, which under the covers will also update the secret.

```go
appDataMap.write(appName, versionName, metricName, summaryMetric) // note that read is on AppData; write is on the data map
```

The above constructs remove the need for cache and simplify code; I wonder if they also eliminate the need for `MetricsSecretStore` data structure also.

abn/grpc/abn.proto
sriumcp: Within the A/B/n service, there is nothing special that is required to fix the track names to a limited set. The only thing that is required is to keep the concepts of track and versions distinct (as described above), so that we document the use of fixed track names.

abn/grpc/abn.proto
kalantar: agreed

abn/watcher/app.go
```diff
@@ -0,0 +1,189 @@
...
+// To do this, the set of track identifiers should be a (small) fixed set, such as "current" and
+// "candidate", that can be mapped to a static routes.
+
+import (
+	"github.com/iter8-tools/iter8/abn/appsummary"
+	"github.com/iter8-tools/iter8/base/log"
+)
+
+// Application is runtime information about the versions of an application
+type Application struct {
```
sriumcp: There seem to be two in-memory data structures for caching app data. One in watcher, and another in appsummary. Could we please coalesce them into one? We could add additional methods to the latter if need be.

sriumcp (changes requested):
Please see a change request

abn/watcher/app.go
kalantar: I considered this. However, the summary object is one that can we directly converted to/from content of secret. The details in the runtime object are not included there. Do you know if skipping some `json` and `yaml` tags will make this work? I hadn't thought of it before.

sriumcp: What are the additional information that the `runtime object` contains which the summary object does not contain? Can those details be constructed simply using additional methods on `ApplicationData` or `ApplicationSummary`? 

The source of truth about applications is the secret. I agree with the need to cache this in-memory for efficient reads. But it is better to keep a single cache, and if need be, restructure the application summary/data types (in order to make both the runtime operations and storage efficient; we can also discuss tradeoffs between these two types of efficiency requirements if need be).

BTW, Kubernetes informers already use a local cache. https://aly.arriqaaq.com/kubernetes-informers/ Perhaps there is no real need for a read-only cache in the first place for application data (assuming we use an informer (or informers) for application data as well)?

kalantar: This does not work.

sriumcp: There are a few different suggestions in my comment above... I don't understand which of those suggestions does not work. Specifically, I am suggesting one or more of the following. Again, the goal is to avoid maintenance of two different caches for the same underlying data.

1. Can those details be constructed simply using additional methods on ApplicationData or ApplicationSummary?
2. It is better to keep a single cache, and if need be, restructure the application summary/data types
3. Kubernetes informers already use a local cache. https://aly.arriqaaq.com/kubernetes-informers/ Perhaps there is no real need for a read-only cache in the first place for application data (assuming we use an informer (or informers) for application data as well).

kalantar: I am thinking more about this and am inclined to think the objects while sharing name and both expressing atttrbiutes of an application (a colletion of versions), they are used quite distinctly. In one use case they are an in memory consturct meant to record observed versions in the cluster. In the second, they are a persistent record of the metrics associated with each version and a history. Here they are read/modified/written (to update) or just read (for assessment). I do not see this as a good fit to merge into a single object. However, I wonder if there is naming that might make their purpose clearer. I tried with Application and ApplicationSummary but the child version objects have the same name.

sriumcp: 1. record observed versions in the cluster
2. record of [...] and a history.

Isn't 2 a superset of 1? Can't 1 be constructed from the recent history (if necessary using additional methods/functions)? Doesn't storing this information twice in-memory twice mean that additional code needs to be written to keep them consistent?

kalantar: Further, when we add the task, I expect the contents of current appsummary to move to base (base needs a notion of summarymetric and the mechanism to read it. I would not expect any of the aspects of the runtime application object ot be present in base.

sriumcp: Also, everything we're discussing so far is about caching for reads. Ultimately, we need to think about write caches also (for metrics), since this is required for scalability. In other words, the history portion of will be used as both read and write cache, and the metrics portion will be used as the write cache.

sriumcp: > I would not expect any of the aspects of the runtime application object to be present in base.

The history information (in particular, current versions and tracks of the app) is the replicated piece.

The idiomatic way to achieve this in Golang is as follows.

```go
package base

type AppSummary struct {...}

func ReadAppSummary(c ClientSet, appName string) AppSummary
```

```go
package abn

type AppSummary base.AppSummary

// more funcs and methods using abn.AppSummary; you still can base.ReadAppSummary here ... to achieve code reuse. For example ... 

func (a *AppSummary) writeMetric(...) error {}
```

kalantar: The runtime and persistent forms of application have been merged together in `abn/application`. A clear part of the model is the part persisted.

abn/watcher/informer.go
```diff
@@ -0,0 +1,82 @@
...
+// MultiInformer holds all of the informers for all namespaces
+type MultiInformer struct {
+	informersByKey             map[string]informers.GenericInformer
+	informerFactroyByNamespace map[string]dynamicinformer.DynamicSharedInformerFactory
+}
+
+// addInformer adds an informer
+func (informer *MultiInformer) addInformer(ns string, gvr schema.GroupVersionResource) {
+	informer.informersByKey[key(ns, gvr)] = informer.informerFactroyByNamespace[ns].ForResource(gvr)
+}
```
kalantar: A result of team review was a recommendation to explore the watch api instead of the informer api.

kalantar: propose we continue to use informers; see comparison https://iter8-tools.slack.com/archives/C01NV4T21HU/p1660598419750459

abn/application/application.go
```diff
@@ -0,0 +1,212 @@
...
+	KEY = "versionData.yaml"
+)
+
+// Application is an application observed in a kubernetes cluster
+type Application struct {
+	// Name is the value of the app.kubernetes.io/name field
+	Name string
+	// Namespace is the namespace where the application was discovered
+	Namespace string
+	// Tracks is map of tracks identified to versions
```
sriumcp: // Tracks is a map from application track name to application version name

abn/application/application.go
```diff
@@ -0,0 +1,212 @@
...
+	Tracks
+	// Versions is list of discovered versions
```
sriumcp: // Versions is a map from application version name to application version object

abn/application/application.go
```diff
@@ -0,0 +1,212 @@
...
+// ApplicationReaderWriter is type used to read/write fromt/to persistent storage
+type ApplicationReaderWriter struct {
+	// Type Kubernetes client
+	Client kubernetes.Interface
+}
+
+// Read reads the application from persistent storage (a Kubernetes secret)
+// - the secret name/namespace is the same as the application
+// - if no application is present in the persistent storage, a new object is created
+func (rw *ApplicationReaderWriter) Read(appName string) (*Application, error) {
```
sriumcp: Isn't it simpler to structure the read and write as follows?

```go
// read application from secret (is a yaml file)
 	rawData, ok := secret.Data[KEY]
 	if !ok {
 		log.Logger.Debug("key missing in backing secret")
 		return a, errors.New("secret does not contain expected key: " + KEY)
 	}
 	err = yaml.Unmarshal(rawData, &a)
```

```go
// write application to secret (is a yaml file)
rawData, err := yaml.Marshal(a)
secret.Data[KEY] = rawData
 	if secret.StringData != nil {
 		secret.StringData[KEY] = string(rawData)
 	}
```

This allows you to mirror the in-memory app to the app data in secret -- in the simplest possible way.

sriumcp: The app data structure is entirely up to you of course (and along with methods/funcs, can be made to yield any information that is needed).

cmd/kdelete.go
```diff
@@ -34,7 +34,7 @@ func newKDeleteCmd(kd *driver.KubeDriver, out io.Writer) *cobra.Command {
 	return cmd
 }
 
-// intialize with the k delete cmd
+// initialize with the k delete cmd
```
sriumcp: These changes need to be merged in the PR.

sriumcp (changes requested):
Initial review comments.

abn/application/application.go
kalantar: reworded

abn/application/application.go
kalantar: reworded

abn/application/application.go
kalantar: The current approach was designed with the principle of minimizing what gets written to the secret. To this end, only the Versions field is written to the secret. The other fields can be determined from the namespace/name of secret and the event history.

Writing all fields is marginally easier -- it is also the case that when there are no events/metrics the read does not initialize these variables. It is necessary to explicitly check and initialize. Doing so on read makes the rest of the code simpler.

I'm inclined to leave it as is for now.

kalantar:
I observe that the current implementation reads/writes to any secret with the appropriate name in the namespace of the application. This means that it might modify an existing secret used for another purpose. An alternative is needed (though the current approach can probably be merged for now). Propose using labels to identify secrets that are related to abn. Perhaps `iter8.tools/abn: <app name>`

abn/application/applicationmap.go
```diff
@@ -0,0 +1,176 @@
...
+	m.mutex.Lock()
+}
+
+// Unlock unlocks the application map
+func (m *ThreadSafeApplicationMap) Unlock() {
+	m.mutex.Unlock()
+}
+
+// Clear the application map
+func (m *ThreadSafeApplicationMap) Clear() {
```
sriumcp: Make clear and add as test helpers.

sriumcp: Implication: locking and unlocking during Get and Write will be handled by the invokers of these methods.

abn/application/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package appliction provides a runtime model of appliations and their versions
```
sriumcp: application(s)

abn/application/test_helper.go
```diff
@@ -0,0 +1,136 @@
...
+	assert.Equal(t, assertion.ready, v.IsReady())
+
+	assert.Len(t, v.Metrics, len(assertion.metrics))
+	assert.NotNil(t, v.Metrics)
+	for m := range v.Metrics {
+		assert.Contains(t, assertion.metrics, m)
+	}
+}
+
+func NoApplications(t *testing.T) {
```
sriumcp: test_helpers -- please make funcs/methods package local, unless they are getting reused outside the package.

abn/watcher/app.go
```diff
@@ -0,0 +1,164 @@
...
+	a, _ := abnapp.Applications.Get(name, false)
+
+	// get the version
+	// if it isn't in the Application this will create an new Version
+	v, _ := a.GetVersion(version, true)
+
+	// set ready to value on watched object, if set
+	// otherwise, use the current readiness value
+	oldReady := v.IsReady()
+	watchedReady := watched.isReady(oldReady)
```
sriumcp:
```go
watchedReady := watched.isReady()
```

abn/watcher/app.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// update track <--> ready version mapping
+	if watchedReady {
+		// log version ready (if it wasn't before)
+		if !oldReady {
```
sriumcp: there is no role for `oldReady`

abn/watcher/app.go
```diff
@@ -0,0 +1,164 @@
...
+			if oldTrack == nil || *oldTrack != watchedTrack {
+				v.Track = &watchedTrack
+				// update a.Tracks
+				a.Tracks[watchedTrack] = version
+			}
+		}
+	} else {
+		// version not ready so if version has track then unmap it
+		// but first check the track to version and remove if mapped to this (not ready) version
+		oldTrack := v.GetTrack()
```
sriumcp: There is no role for `oldTrack`.

abn/watcher/app.go
```diff
@@ -0,0 +1,164 @@
...
+	log.Logger.Trace("Update called")
+	defer log.Logger.Trace("Update completed")
+
+	Add(watched)
+}
+
+// Delete updates the apps map using information from a deleted object
+// Note that we are not object counting which means we will never actually remove a version
+// from an application or an application from the syste
+func Delete(watched WatchedObject) {
```
sriumcp: No more explicit ready field. The user should not add a track label unless a version object is ready (track implies ready version).

abn/watcher/app.go
sriumcp: No role for `old*` (including tracks).

abn/watcher/config.go
```diff
@@ -0,0 +1,52 @@
...
+	"io/ioutil"
+
+	"github.com/iter8-tools/iter8/base/log"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"sigs.k8s.io/yaml"
+)
+
+// Config is A/B(/n) serivce configuration. The configuration identifies the resources to watch and in which namespaces.
+type Config struct {
```
sriumcp: Consider creating package local types and funcs.

sriumcp: Perhaps move it to `callee` package.

driver/kubedriver.go
```diff
@@ -91,6 +95,12 @@ func (kd *KubeDriver) InitKube() error {
 			log.Logger.WithStackTrace(err.Error()).Error(e)
 			return e
 		}
+		kd.DynamicClient, err = dynamic.NewForConfig(restConfig)
```
sriumcp: Is kubedriver getting used in abn experiment service? Is this change required?

abn/watcher/app.go
```diff
@@ -0,0 +1,164 @@
...
+
+import (
+	abnapp "github.com/iter8-tools/iter8/abn/application"
+	"github.com/iter8-tools/iter8/base/log"
+)
+
+// Add updates the apps map using information from a newly added object
+// If the observed object does not have a name (app.kubernetes.io/name label)
+// or version (app.kubenetes.io/version), it is ignored.
+func Add(watched WatchedObject) {
```
sriumcp: These are package local funcs.

abn/watcher/watchedobject.go
```diff
@@ -0,0 +1,77 @@
...
+	annotations := wo.Obj.GetAnnotations()
+	track, ok := annotations[TRACK_ANNOTATION]
+	if !ok {
+		return ""
+	}
+	return track
+}
+
+// isReady determines if watched object indicates readiness of version (as indicated by READY_ANNOTATION annotatation)
+func (wo WatchedObject) isReady(currentlyReady bool) bool {
```
sriumcp: isReady is going away.

abn/grpc/lookup.go
```diff
@@ -0,0 +1,61 @@
+package grpc
```
sriumcp: perhaps `package abn`

abn/service.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil || track == nil {
+		return nil, err
+	}
+
+	return &pb.Session{
+		Track: *track,
+	}, err
+}
+
+func (server *abnServer) WriteMetric(ctx context.Context, metricMsg *pb.MetricValue) (*emptypb.Empty, error) {
```
sriumcp: Make `WriteMetric` as clean as possible, like `Lookup`. Perhaps split the function implementation as in Lookup?

abn/service.go
```diff
@@ -0,0 +1,150 @@
...
+
+type abnServer struct {
+	Driver *driver.KubeDriver
+	pb.UnimplementedABNServer
+}
+
+// Lookup identifies a track that should be used for a given user
+// This method is exposed to gRPC clients
+func (server *abnServer) Lookup(ctx context.Context, appMsg *pb.Application) (*pb.Session, error) {
+	abnapp.Applications.Lock()
```
sriumcp: For reads, use `RLock()` and not `Locks()`.

sriumcp: We want `Rlocks` and `Locks` at per application level (for application-level operations). For map-level operations, use the map level locks.

abn/service.go
sriumcp: Hide details.

abn/service.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		log.Logger.Warn("Unable to parse metric value ", metricMsg.GetValue())
+		return &emptypb.Empty{}, nil
+	}
+
+	m, _ := v.GetMetric(metricMsg.GetName(), true)
+	m.Add(value)
+
+	// persist updated metric
+	abnapp.Applications.Write(a)
```
sriumcp: Update this secret only if the last update to this secret was 60 seconds ago.

sriumcp: Create a BatchedWrite(?) method for use in WriteMetrics. 
***
When the application object gets too big, we drop untracked versions.
***
There is a limit on number of metrics-per-version.

sriumcp (changes requested):
Changes requested.

abn/service.go
kalantar: done

abn/grpc/lookup.go
kalantar: moved

abn/service.go
kalantar: added control per application in addition to the map
distinguished between read and write locks

abn/application/doc.go
kalantar: done

abn/application/applicationmap.go
kalantar: Left Add() as a map method; refactored all adds of new maps to use this method
Moved Clear() to test_helper since used only in tests
Refactored locks to be application specific as well as map specific. Map locks now in Get/Add/Clear; no longer assume caller calls them.
This seemed simpler.

abn/service.go
kalantar: Probably should make the timeout value be user configurable at install time.

abn/watcher/app.go
kalantar: modified based on decision that there is EXACTLY one object that defines a version of an application. In light of this, there is no need for `oldReady`.

abn/application/test_helper.go
kalantar: number of exported helper functions reduced to 1.

abn/watcher/app.go
kalantar: ready field removed

abn/watcher/watchedobject.go
kalantar: redefined

abn/watcher/config.go
kalantar: refactored

kalantar: found I had to keep the fields exported (capitalized). Otherwise yaml.Unmarshal does not work. See https://github.com/golang/go/issues/8598

abn/watcher/app.go
kalantar: agreed. renamed.

driver/kubedriver.go
kalantar: It is using this right now.  It is debatable if it should.
Currently, the`iter8 abn` is configured with the global `KubeDriver` much as all `iter8 k` commands are. This is passed to the abn service `Start` method and used by the service to access the cluster. 

`KubeDriver` is not intended for this purpose (ie, general access to a cluster). It is intended only for read/wrte experiments.

To change this, a general cluster client is needed. This is needed not only by  abn service but also by things like the ready task (which is using a copy of the KubeDriver -- still called KubeDriver though). This should be used instead from the `iter8 abn` command. The KubeDriver can be written to extend this more general client.

In preparing the work for the abn task, I have prototyped this refacoring (that task also needs this). However, it could be added now.  The apparoch was to add a package `github.com/iter8-tools/iter8/base/k8sdriver` containing the necessary definitions. This is used by readiness, the abn task and the abn service. I don't recall if KubeDriver was refactored to extend this in the prototype.

abn/service.go
kalantar: How do I compute the "size" of a secret?

kalantar: set maximum length of byte array of content of application data (converted to yaml) to 750,000

driver/kubedriver.go
sriumcp: Given that requirements are still evolving as we develop this capability, perhaps one approach could be to...
1. Leave the Kubedriver package and base Kubedriver functions alone (and not use them in abn and autox)
2. Create any Kubernetes client functionality needed entirely within the abn package (of course any code that you need can be copied from base or kubedriver packages)
3. Create any Kubernetes client functionality needed entirely within the autox package (of course any code that you need can be copied from base or kubedriver packages or abn package)

Once we have an MVP release of both autox and abn, we can reassess exactly how the kubernetes client functionality should be structured across base-for-readiness/kubedriver-for-experiment launch/kubeclient-for-abn/kubeclient-for-autox.

Of course, they all have some level of overlap (for e.g., they all need to access the cluster). But each of them have a distinct purpose.

sriumcp: The main reason for keeping `abn` and `autox` in the same repo as `iter8` is the reuse of CI (and images). 
***
That said, we can also look into other code reuse eventually, after we have the MVP release.

kalantar: We also need this capability (access to kubernetes) in tasks. readiness and read abn metrics tasks both need it. We've already repeated it there (in base) but we already know it should be changed. I would suggest partially fixing now (no extending KubeDriver based on it.

I'm inclined to think 3 copies of code is too many.

kalantar: I was going to say that it is clear what is needed. I know for the tasks and for abn. But I guess autox may need more to support helm. KubeDriver has this. Not sure it is the same package though. Maybe copy is ok

kalantar: I'm a little worried that we will never get to refactoring this.

sriumcp: There is the risk of of rushing through the refactoring (without fully understanding the usage scenarios), and the risk of never refactoring this and replicating code. 
***
I think the latter is preferable to the former. At least, bugs in one component will not affect others. Besides, we are mixing design with prototyping extensively (in both ABn and autox). This is perfectly fine, as long as we isolate these changes, and not let it spill over into current Iter8 functions. What we have for ABn and autox is not just an MVP implementation, but also an MVP design, which means our next version might very well change things for these packages.
***
Could we  please create an issue (if one doesn't already exist), and revisit after the MVPs.

kalantar: done

abn/application/application.go
```diff
@@ -0,0 +1,117 @@
...
+type Application struct {
+	// Name is of the form namespace/name where the name is the value of the label app.kubernetes.io/name
+	Name string
+	// Tracks is map from application track identifier to version name
+	Tracks
+	// Versions is a map of versions name to version data
+	Versions
+}
+
+// Versions is type of version name to versions map
```
sriumcp: // Versions maps the name of the version to the version object

abn/application/application.go
```diff
@@ -0,0 +1,117 @@
...
+type Versions map[string]*Version
+
+// Tracks is map of track identifiers to version names
+type Tracks map[string]string
+
+// ApplicationReaderWriter is type used to read/write fromt/to persistent storage
```
sriumcp: spell check

abn/application/application.go
```diff
@@ -0,0 +1,117 @@
...
+type ApplicationReaderWriter struct {
+	// Client is the Kubernetes client to use to read/write secrets
+	Client kubernetes.Interface
+}
+
+// NewApplication returns a new (empty) Application for a namespace/name label
+func NewApplication(application string) *Application {
```
sriumcp: NewApplication becomes newApplication.

abn/application/application.go
```diff
@@ -0,0 +1,117 @@
...
+	return &Application{
+		Name:     application,
+		Versions: Versions{},
+		Tracks:   Tracks{},
+	}
+}
+
+// GetNameFromKey returns the name from a key of the form "namespace/name"
+func GetNameFromKey(applicationKey string) string {
```
sriumcp: Should `Get*` functions be exported?

abn/application/application.go
```diff
@@ -0,0 +1,117 @@
...
+	if len(names) > 1 {
+		namespace, name = names[0], names[1]
+	} else {
+		namespace, name = "default", names[0]
+	}
+
+	return namespace, name
+}
+
+// GetVersion returns the Version identified by version
```
sriumcp: returns the `version object` corresponding to a given `version name`.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+// initalize global variables
+func init() {
+	Applications = ThreadSafeApplicationMap{
+		apps:           map[string]*Application{},
+		mutexes:        map[string]*sync.RWMutex{},
+		lastWriteTimes: map[string]*time.Time{},
+		rw:             k8sclient.NewKubeClient(cli.New()),
+	}
+	Applications.rw.Initialize()
+	BatchWriteInterval = time.Duration(60 * time.Second)
```
sriumcp: `60` seconds should be a constant.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+	maxApplicationDataBytes = 750000 // a secret's maximum size is 1MB
```
sriumcp: constant.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+
+	a, err := m.Read(application)
+	m.Add(a)
+	return a, err
+}
+
+// Read reads the application from persistent storage (a Kubernetes secret)
+// - the secret name/namespace is the same as the application
+// - if no application is present in the persistent storage, a new object is created
+func (m *ThreadSafeApplicationMap) Read(application string) (*Application, error) {
```
sriumcp: Should this be externalized?

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// read data from secret (is a yaml file)
+	rawData, ok := secret.Data[KEY]
+	if !ok {
+		log.Logger.Debug("key missing in backing secret")
+		return NewApplication(application), errors.New("secret does not contain expected key: " + KEY)
+	}
+
+	// err = yaml.Unmarshal(rawData, &a.Versions)
+	a := &Application{}
```
sriumcp, line 151: Please absorb 
```go
	// initialize a.Versions if not already
 	if a.Versions == nil {
 		a.Versions = Versions{}
 	}
 	for _, v := range a.Versions {
 		if v.Metrics == nil {
 			v.Metrics = map[string]*SummaryMetric{}
 		}
 	}
```

into a custom golang JSON unmarshaler for Application object.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+	defer log.Logger.Trace("Write completed")
+	var secret *corev1.Secret
+
+	// marshal to byte array
+	rawData, err := yaml.Marshal(a)
+	if err != nil {
+		return err
+	}
+
+	if len(rawData) > maxApplicationDataBytes {
```
sriumcp: Check if it makes sense to write a custom marshaler... and if so, whether it needs to be a JSON marshal or YAML marshal.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+		deleteUntrackedVersions(a)
+		rawData, err = yaml.Marshal(a)
+		if err != nil {
+			return err
+		}
+	}
+
+	if len(rawData) > maxApplicationDataBytes {
```
sriumcp: Describe the size limit for tracked versions and their metrics, clearly. When the limit is breached, we need to signal errors that users can easily take note of.
***
Starting point could just be `iter8 abn logs`.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	// update last write time for application
+	now := time.Now()
+	m.lastWriteTimes[a.Name] = &now
+	return nil
+}
+
+// BatchedWrite writes the Application to persistent storage only if the previous write was more than TIME ago
+func (m *ThreadSafeApplicationMap) BatchedWrite(a *Application) error {
```
sriumcp, line 234: Issue 1: Periodically flush application
Issue 2: Write users-per-track/users-per-version information

abn/application/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package appliction provides a runtime model of applications and their versions
```
sriumcp: spell check

abn/application/test_helper.go
```diff
@@ -0,0 +1,100 @@
...
+
+	assert.Len(t, v.Metrics, len(assertion.metrics))
+	assert.NotNil(t, v.Metrics)
+	for m := range v.Metrics {
+		assert.Contains(t, assertion.metrics, m)
+	}
+}
+
+// Clear the application map
+func (m *ThreadSafeApplicationMap) Clear() {
```
sriumcp, line 103: Consider an `internal` package for `Clear` and `NumApplications`.

driver/kubedriver.go
```diff
@@ -53,8 +54,10 @@ const (
 type KubeDriver struct {
 	// EnvSettings provides generic Kubernetes and Helm options
 	*cli.EnvSettings
-	// Clientset enables interaction with a Kubernetes cluster
+	// Clientset enables interaction with a Kubernetes cluster using structured types
```
sriumcp: To be undone.

sriumcp (changes requested):

abn/application/application.go
kalantar: fixed

abn/application/application.go
kalantar: fixed.

abn/application/application.go
kalantar: done

abn/application/application.go
kalantar: rerworded

abn/application/applicationmap.go
kalantar: done

abn/application/applicationmap.go
kalantar: done

abn/application/application.go
kalantar: made local and moved to `applicationmap.go` where they are actually used. Refactored test cases that were using these so they are no longer necessary.

abn/application/applicationmap.go
kalantar: made local

abn/application/doc.go
kalantar: corrected

abn/application/test_helper.go
kalantar: An internal package does not seem like the right solution to our challenge. The external functions are accessing package local fields. It is not possible to move them to an internal package without moving the objects that contain those fields. One would need to move pretty much everything and for each item to be externalized a wrapper written to make it public.

driver/kubedriver.go
kalantar: undid all changes to `driver` package

abn/application/applicationmap.go
kalantar: done

abn/application/applicationmap.go
kalantar: done by defining a MarshalJSON method

abn/application/applicationmap.go
kalantar: implemented 1; created [issue](https://github.com/iter8-tools/iter8/issues/1330) for 2

abn/application/applicationmap.go
kalantar: created [issue](https://github.com/iter8-tools/iter8/issues/1331)

abn/application/application.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Versions is a map of the version name to a version object
+type Versions map[string]*Version
+
+// Tracks is map of track identifiers to version names
+type Tracks map[string]string
+
+// ApplicationReaderWriter is used to read/write from/to a Kubernetes secret
+type ApplicationReaderWriter struct {
```
sriumcp: This type goes away.

abn/application/applicationmap.go
```diff
@@ -0,0 +1,309 @@
...
+	} else {
+		namespace, name = "default", names[0]
+	}
+
+	return namespace, name
+}
+
+func (m *ThreadSafeApplicationMap) PeriodicApplicationsFlush() {
+	for {
+		time.Sleep(5 * BatchWriteInterval)
```
sriumcp: Make a constant out of 5. e.g., `FlushMultiplier` or `FlushInterval`.

abn/k8sclient/k8sclient.go
```diff
@@ -0,0 +1,67 @@
...
+	dynamicClient dynamic.Interface
+}
+
+// initKube initializes the Kubernetes clientset
+func (kd *KubeClient) Initialize() (err error) {
+	if kd.dynamicClient == nil {
+
+		// get rest config
+		// see https://pkg.go.dev/k8s.io/client-go/tools/clientcmd
+		loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
```
sriumcp: Could we please use this to get the config: https://github.com/iter8-tools/iter8/blob/4a2ae79984d5710b0b5ec7a13666c90e6b815eca/driver/kubedriver.go#L78

sriumcp: Enables local run in addition to in-cluster.

abn/lookup.go
```diff
@@ -0,0 +1,64 @@
+package abn
```
sriumcp: Put the files under  `abn` into something like a `core` package under abn folder?

There may not be an abn package at all.

abn/lookup.go
```diff
@@ -0,0 +1,64 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"hash/crc32"
+
+	abnapp "github.com/iter8-tools/iter8/abn/application"
+	"github.com/iter8-tools/iter8/base/log"
+)
+
+func rendezvousGet(a *abnapp.Application, key string) string {
```
sriumcp: Comments on these funcs would be very helpful.

sriumcp: // key here is a user id.

abn/lookup.go
```diff
@@ -0,0 +1,64 @@
...
+	var maxScore uint32
+	var maxNode []byte
```
sriumcp: // versionBytes instead of maxNode

sriumcp: // rendezvousGet returns a consistent track for a given user and application. The track is chosen uniformly at random from available tracks, but this selection is sticky for a given app and user (i.e., randomness is across users, but not across invocations for a given user/app).

sriumcp: // when the app data structure changes, the random hashing used to map users onto tracks should also change.

sriumcp (changes requested):

abn/application/application.go
kalantar: removed

abn/application/applicationmap.go
kalantar: done

abn/k8sclient/k8sclient.go
kalantar: reverted

abn/lookup.go
kalantar: comments added and variables renamed to make it clearer

kalantar: done

abn/lookup.go
kalantar: renamed variables and added documentation to explain implementation

abn/lookup.go
kalantar: moved files in abn/ to abn/core (and renamed package)

abn/core/service.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	stopCh := make(chan struct{})
+
+	// set up resource watching as defined by config
+	c := readConfig(abnConfigFile)
+	w := watcher.NewIter8Watcher(c.Resources, c.Namespaces)
+	go w.Start(stopCh)
+
+	abnapp.Applications.PeriodicApplicationsFlush(stopCh)
```
sriumcp:
```go
// this launches a go routine under the covers
```

abn/watcher/app.go
```diff
@@ -0,0 +1,140 @@
...
+		log.Logger.Debug("no version found")
+		return
+	}
+
+	// the watched object is the object that defines the version of the application
+
+	// check if we know about this application
+	// first check if in memory then read from persistent store if not found
+	// if it isn't in persistent store, the read will return an initalized Application
+	a, _ := abnapp.Applications.Get(name, false)
```
sriumcp: Pending analysis of race conditions. Are there any race conditions? Are there any dead locks?

sriumcp: Exactly what are the operations for which atomicity is guaranteed? And can they be interleaved in any arbitrary manner -- without compromising eventual consistency.

abn/watcher/informer.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/iter8-tools/iter8/abn/k8sclient"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+)
+
+type Iter8Watcher struct {
```
sriumcp: Is this used by an external package? If not, please remove capitalization.

sriumcp (changes requested):

abn/core/lookup.go
```diff
@@ -0,0 +1,79 @@
...
+// associated with an application.
+// We want to always return the same track for the same user so long as the
+// application remains unchanged -- there are no change in the set of versions
+// and no change to the track mapping.
+// To do this, we hash the combination of user and version. We don't use the track identifier
+// because the track identifier is associated with multiple versions over time; we do not
+// require a fixed mapping when this mapping changes.
+// We select the version, user pair with the largest hash value ("score").
+// Inspired by https://github.com/tysonmote/rendezvous/blob/master/rendezvous.go
+func rendezvousGet(a *abnapp.Application, user string) string {
```
sriumcp, line 49: Something along the following lines for consistent hashing of users to tracks.
```go
// ctvus stands for consistently hashable track version user triples

type ctvus []string

func (a ctvus) Len() int           { return len(a) }
func (a ctvus) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
func (a ctvus) Less(i, j int) bool {
  var h maphash.Hash
  var g maphash.Hash
  h.WriteString(a[i])
  g.WriteString(a[j])
  return h.Sum64() < g.Sum64()
}

func (a *Application) lookUp(user string) string {
  tvus := ctvus{} // track version user triples
  for t, v := range a.Tracks { // t is track; v is version
    tvus = append(tvs, t + ":" + v + ":" + user) // track:version:user triples
  }
  sort.Sort(tvus)
  s := strings.Split(tvus[0], ":")
  return s[0]
}
```

sriumcp (changes requested):
lookup func

abn/core/lookup.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+	return maxTrack
+}
+
+// hasher is a 32 bit hash function
+var hasher = crc32.New(crc32.MakeTable(crc32.Castagnoli))
+
+// hash computes the score for a version, user combination
+func hash(version, user []byte) uint32 {
```
sriumcp: Perhaps ... 
```go
// hash computes the score for a version, user combination
 func hash(version string, user string) uint32 {
        hasher := maphash.Hash
 	hasher.WriteString(user)
 	hasher.WriteString(version)
 	return hasher.Sum32()
 }
```

This eliminates need for byte operations.

We can perhaps replace `bytes.Compare(versionBytes, maxVersion)` with `v > maxVersion`

sriumcp (commented):
Please see suggested change for lookup.

abn/watcher/informer.go
kalantar: done

abn/watcher/app.go
kalantar: fixed potential issue in Applications.Add()

abn/watcher/informer.go
```diff
@@ -0,0 +1,46 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+)
+
+type iter8Watcher struct {
+	factories map[string]dynamicinformer.DynamicSharedInformerFactory
+}
+
```
sriumcp, line 16:
```go
var addFunc := func(obj interface{}) {
  addObject(watchedObject{Obj: obj.(*unstructured.Unstructured)})
}
```

abn/watcher/informer.go
```diff
@@ -0,0 +1,46 @@
...
+	w := &iter8Watcher{
+		factories: map[string]dynamicinformer.DynamicSharedInformerFactory{},
+	}
+	// for each namespace, resource type configure Informer
+	for _, ns := range namespaces {
+		w.factories[ns] = dynamicinformer.NewFilteredDynamicSharedInformerFactory(k8sclient.Client.Dynamic(), 0, ns, nil)
+		for _, gvr := range resourceTypes {
+			informer := w.factories[ns].ForResource(gvr)
+			informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
```
sriumcp: `AddFunc: addFunc`

abn/watcher/informer_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/iter8-tools/iter8/abn/k8sclient"
+	"github.com/stretchr/testify/assert"
+
+	"helm.sh/helm/v3/pkg/cli"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestNewInformer(t *testing.T) {
+	k8sclient.Client = *k8sclient.NewFakeKubeClient(cli.New())
```
sriumcp, line 306:
```go
numAdditions := 0
addFunc = func(obj) {
  numAdditions += 1
}
...
assert.DelayedEqual(t, 1, numAdditions, 30s)
```

abn/core/lookup.go
kalantar: done

abn/watcher/informer_test.go
kalantar: added new informer test that relies on domain specific behavior to verify triggers

abn/core/service.go
kalantar: done

abn/application/application.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// Application is an application observed in a kubernetes cluster
+type Application struct {
+	// Name is of the form namespace/name where the name is the value of the label app.kubernetes.io/name
+	Name string `json:"name" yaml:"name"`
+	// Tracks is map from application track identifier to version name
+	Tracks `json:"tracks" yaml:"tracks"`
+	// Versions is a map of versions name to version data
+	Versions `json:"versions" yaml:"versions"`
```
sriumcp, line 21: // Map each version to its metrics (version data = metrics)

abn/application/summary_metric.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"math"
+)
+
+// SummaryMetric is summary metric type
+type SummaryMetric [5]float64
+
+const (
+	COUNT_IDX = 0
```
sriumcp: `countIdx` rather than `COUNT_IDX`.

abn/grpc/abn.proto
```diff
@@ -0,0 +1,63 @@
...
+  // Should be called for each request (transaction).
+  rpc Lookup(Application) returns(Session) {}
+
+  // Write a metric value to metrics database.
+  // The metric value is explicitly associated with a list of transactions that contributed to its computation.
+  // The user is expected to identify these transactions.
+  rpc WriteMetric(MetricValue) returns (google.protobuf.Empty) {}
+
+  // Get metrics for an application
+  rpc GetMetrics(MetricRequest) returns (ApplicationMetrics) {}
```
sriumcp: `GetApplication` or  `GetApplicationData`.

sriumcp (changes requested):
Very minor changes.

abn/application/application.go
kalantar: comment improved

abn/application/summary_metric.go
kalantar: done

abn/grpc/abn.proto
kalantar: renamed

abn/core/service.go
```diff
@@ -0,0 +1,123 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"helm.sh/helm/v3/pkg/cli"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+const (
+	// Name of environment variable with file path to configuration yaml file
+	WATCHER_CONFIG_ENV = "WATCHER_CONFIG"
```
sriumcp: `watcherConfigEnv`

abn/application/application.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/iter8-tools/iter8/base/log"
+)
+
+// Application is an application observed in a kubernetes cluster
+type Application struct {
+	// Name is of the form namespace/name where the name is the value of the label app.kubernetes.io/name
+	Name string `json:"name" yaml:"name"`
```
sriumcp, line 17: `name string`, `tracks` and `Versions: `json:"versions" yaml:"versions"`

This ensures only version data is ever serialized for the app.

abn/watcher/handler.go
```diff
@@ -0,0 +1,109 @@
...
+		if o.isReady() {
+			track := o.getTrack()
+			if track != "" {
+				a.Tracks[track] = version
+			}
+		}
+	}
+	abnapp.Applications.Unlock(application)
+
+	abnapp.Applications.Write(a)
```
sriumcp: `abnapp.Application.Write(a)` can be taken out.

abn/watcher/handler.go
```diff
@@ -0,0 +1,109 @@
...
+// getApplicationObjects gets all the objects related to the application based on label app.kubernetes.io/name
+func getApplicationObjects(namespace, name string, gvrs []schema.GroupVersionResource, informerFactories map[string]dynamicinformer.DynamicSharedInformerFactory) []watchedObject {
+	// define selector
+	selector := labels.NewSelector()
+	reqSpec := []struct {
+		key  string
+		op   selection.Operator
+		vals []string
+	}{
+		{key: ITER8_LABEL, op: selection.Equals, vals: []string{"true"}},
```
sriumcp: `iter8Label` instead of `ITER8_LABEL` (and so on).

abn/watcher/informer_test.go
```diff
@@ -0,0 +1,386 @@
...
+						s.application == "" ||
+						s.version == "" {
+						r = r && abnapp.NumApplications(t, 0)
+						return r
+					}
+
+					// otherwise application created
+
+					a, err := abnapp.Applications.Get(s.namespace + "/" + s.application)
+					r = r && assert.NoError(t, err)
```
sriumcp, line 64:
```go
if !assert.NoError(t, err) {return false}
```

sriumcp: Or try `require.NoError(t, err)`

abn/watcher/informer_test.go
```diff
@@ -0,0 +1,386 @@
...
+				Resource: "deployments",
+			}
+
+			done := make(chan struct{})
+			defer close(done)
+			setup(t, gvr, s.namespace, done)
+
+			createObject(t, gvr, s.iter8related, s.namespace, s.application, s.version, s.track, s.ready)
+
+			assert.Eventually(
```
sriumcp, line 49: Will `require.Eventually` work?

abn/watcher/watchedobject.go
```diff
@@ -0,0 +1,77 @@
...
+// watchedobject.go - methods to read fields from an unstructured Kubernetes object
+
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const (
+	NAME_LABEL       = "app.kubernetes.io/name"
```
sriumcp: `nameLabel`.

abn/watcher/watchedobject.go
```diff
@@ -0,0 +1,77 @@
...
+	VERSION_LABEL    = "app.kubernetes.io/version"
+	READY_ANNOTATION = "iter8.tools/ready"
+	TRACK_ANNOTATION = "iter8.tools/track"
```
sriumcp: `trackLabel`. and no `readyAnnotation`.

sriumcp (changes requested):
Minor requests for changes.

abn/core/service.go
kalantar: done

sriumcp (approved):
lgtm

PR #1314, opened 2022-08-08
Slack notification

Alan-Cha:


base/custom_metrics.go
```diff
@@ -251,10 +252,6 @@ func (t *customMetricsTask) run(exp *Experiment) error {
 		return err
 	}
 
-	if exp.driver == nil {
```
Alan-Cha, line 254: I think this was left over after `collect-metrics-database` task was restructured to `custom_metrics`. We do not use `exp.driver` here anymore.

base/custom_metrics.go
```diff
@@ -251,10 +252,6 @@ func (t *customMetricsTask) run(exp *Experiment) error {
...
-		return errors.New("no driver was provided for collect-metrics-database task")
-	}
-
```
Alan-Cha, line 257: I think this was left over after `collect-metrics-database` task was restructured to `custom_metrics`. We do not use `exp.driver` here anymore.

Alan-Cha:
Slack notification sample:
Screen Shot 2022-08-09 at 3 45 35 PM
Currently, this is just outputting the experiment, which is too long to be practical. We can easily create new payload templates so that something more reasonable can be outputted ([current template](https://github.com/Alan-Cha/iter8/blob/slack-notification/charts/iter8/templates/_payload-slack.tpl)).

In previous [commits](https://github.com/Alan-Cha/iter8/commit/5372830d6260696914174f46988f833e98e38b6a), I have explored what else I could do with Slack notifications. Something to note is that code blocks are limited to 3000 characters, which is too short to output the entirety of the experiment. I also looked into uploading files, which would allow us to preview the file in the notification (instead of showing the entirety at its full length) but that will require additional permissions.

```bash
iter8 launch \
--noDownload \
--set "tasks={http,assess,slack}" \
--set http.url=http://127.0.0.1/get \
--set assess.SLOs.upper.http/latency-mean=50 \
--set assess.SLOs.upper.http/error-count=0 \
--set slack.url="https://hooks.slack.com/services/.../..." \
--set slack.payloadTemplateURL="https://raw.githubusercontent.com/Alan-Cha/iter8/slack-notification/charts/iter8/templates/_payload-slack.tpl" \
--set runner=job
```

***

GitHub notification sample:
Screen Shot 2022-08-09 at 3 11 03 PM

As you can see, the workflow also has access to the `client_payload`, which means that the user can use information from the experiment for the workflow tasks.

```bash
iter8 launch \
--noDownload \
--set "tasks={http,assess,github}" \
--set http.url=http://127.0.0.1/get \
--set assess.SLOs.upper.http/latency-mean=50 \
--set assess.SLOs.upper.http/error-count=0 \
--set github.owner=Alan-Cha \
--set github.repo=iter8-notify \
--set github.token=... \
--set runner=job
```

Alan-Cha:
I think we should also consider allowing the user to pass additional payload template values through the `notify` task inputs. This would allow the user to have greater control over the payload template. 

One use case would be to allow the use to change the `repository_dispatch` event type in the GitHub notify task workflow ([see here](https://github.com/Alan-Cha/iter8-notify/blob/ddc4d5a10cb5d0a46c43680de9228d639094f934/.github/workflows/iter8-notify.yml#L4)). A user could then control which workflow will run with how they see fit.

Alan-Cha:
I am not sure where the payload templates should live. Currently, they are with all the other templates.

base/notify.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// NumLoops is the current loop of the experiment
+	NumLoops int `json:"numLoops" yaml:"numLoops"`
+
+	// Experiment is the experiment struct
+	Experiment *Experiment `json:"experiment" yaml:"experiment"`
+}
+
+// NotifyPayloadTemplateValues contains the report as well as the report in some other formats
+type NotifyPayloadTemplateValues struct {
```
sriumcp: I wonder if this type is required at all. https://github.com/iter8-tools/iter8/blob/50f45970a5ccad4f616df80f8aba7d40999ba6c4/base/custom_metrics.go#L221 -- already includes Sprig's (Helm's) [type conversion template functions](https://helm.sh/docs/chart_template_guide/function_list/#type-conversion-functions)... and this in turn includes JSON conversion (with and without escaping).

base/notify.go
```diff
@@ -0,0 +1,247 @@
...
+	}, nil
+}
+
+// getPayload fetches the payload template from the PayloadTemplateURL and
+// executes it with values from getPayloadTemplateValues()
+func (t *notifyTask) getPayload(exp *Experiment) (string, error) {
+	if t.With.PayloadTemplateURL != "" {
+		template, err := getProviderTemplate(t.With.PayloadTemplateURL)
+		if err != nil {
+			log.Logger.Error("could not get payload template")
```
sriumcp: This log here is not needed. The original `getProviderTemplate` should already do the logging.

base/custom_metrics.go
```diff
@@ -232,6 +232,7 @@ func getProviderTemplate(providerURL string) (*template.Template, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
sriumcp: We should rename `getProviderTemplate` to`getTextTemplateFromURL` and move it to [`util.go`](https://github.com/iter8-tools/iter8/blob/master/base/util.go) so that both custom metrics and notify tasks can (re)use this function.

base/notify.go
```diff
@@ -0,0 +1,247 @@
...
+
+	marshalledReport, err := json.MarshalIndent(report, "", "  ")
+	if err != nil {
+		log.Logger.Error("could not JSON marshall report")
+		return nil, err
+	}
+
+	stringReport := string(marshalledReport)
+
+	// escape double quotes, tabs, and new lines
```
sriumcp: We really don't want anything special here ... but if special functionality is indeed required, `getTextTemplateFromURL` can be modified so that additional template functions (e.g., for escaping) is introduced as part of the template func map.

sriumcp: The idea is to get rid of NotifyPayloadTemplateValues struct altogether and work only with Report.

charts/iter8/templates/_payload-github.tpl
```diff
@@ -0,0 +1,4 @@
+{
+  "event_type": "iter8",
+  "client_payload": {{ .JSONStringReport | indent 2 | trim}}
```
sriumcp: Instead of `JSONStringReport`, we could instead try `. | toJson` (or prettyJson, or rawJson). These funcs are already part of the current provider template func map.

sriumcp (changes requested):
Please see review comments on template func maps.

charts/iter8/templates/_payload-slack.tpl
```diff
@@ -0,0 +1,3 @@
+{
```
sriumcp: Payload templates can be hosted anywhere. Perhaps we should simply create a gist for the (sample slack/GitHub) payload templates.

charts/iter8/templates/_task-slack.tpl
```diff
@@ -0,0 +1,19 @@
+{{- define "task.slack" -}}
```
sriumcp: Task templates go here https://github.com/iter8-tools/hub/tree/main/charts/iter8lib/templates

charts/iter8/templates/_task-github.tpl
```diff
@@ -0,0 +1,25 @@
+{{- define "task.github" -}}
```
sriumcp: Task templates go here https://github.com/iter8-tools/hub/tree/main/charts/iter8lib/templates

sriumcp (changes requested):
Please see further comments.

charts/iter8/templates/_task-slack.tpl
```diff
@@ -0,0 +1,19 @@
...
+{{- /* Validate values */ -}}
+{{- if not . }}
+{{- fail "slack notify values object is nil" }}
+{{- end }}
+{{- if not .url }}
+{{- fail "please set a value for the url parameter" }}
+{{- end }}
+{{- if not .payloadTemplateURL }}
+{{- fail "please set a value for the payloadTemplateURL parameter" }}
+{{- end }}
```
kalantar: you are requiring this but have a default below. not required?

charts/iter8/templates/_task-github.tpl
```diff
@@ -0,0 +1,25 @@
...
+{{- end }}
+{{- if not .owner }}
+{{- fail "please set a value for the owner parameter" }}
+{{- end }}
+{{- if not .repo }}
+{{- fail "please set a value for the repo parameter" }}
+{{- end }}
+{{- if not .token }}
+{{- fail "please set a value for the token parameter" }}
+{{- end }}
```
kalantar: I'm curious if softfailure true means we should not fail when there are configuration problems?  Or just runtime failures.

base/notify.go
```diff
@@ -0,0 +1,247 @@
...
+		} else {
+			t.With.Method = "GET"
+		}
+	}
+}
+
+// validate task inputs
+func (t *notifyTask) validateInputs() error {
+	return nil
+}
```
kalantar: I feel like the url and PayloadTemplateURL should be verified as existing. They are set in current helm templates but this validation should cover other possibilities.

base/notify.go
```diff
@@ -0,0 +1,247 @@
...
+		return nil
+	}
+	defer resp.Body.Close()
+
+	// read response responseBody
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		log.Logger.Error("could not read response body from Slack notification request", err)
+		return nil
+	}
```
kalantar: should we also be looking at resp.StatusCode() and failing if not 200/202 ?

base/notify.go
Alan-Cha: I see... I will give this a try and report back.

base/notify.go
Alan-Cha: Do you have any advice on how to determine if we should log something and when we should just return the error?

base/notify.go
Alan-Cha: Okay, we can do that.

base/notify.go
Alan-Cha: Yeah, I think that is the point of `softFailure` which I forgot to implement. I will do that.

charts/iter8/templates/_payload-slack.tpl
Alan-Cha: Okay, sounds good to me.

charts/iter8/templates/_task-github.tpl
Alan-Cha: I think `softFailure` `true` should mean that the experiment fails when there are configuration problems but not when there are runtime failures.

charts/iter8/templates/_task-slack.tpl
Alan-Cha: Sorry, that was a mistake! Last minute addition.

base/notify.go
sriumcp: The first place where an error is discovered/created in Iter8 -- that's where logging needs to happen. For example, `getProvidereTemplate` already should log an error message (since it `knows` that it is returning an error), but if you Log that again in `notify.go`, the same log will appear twice.

base/notify.go
sriumcp: `PayloadTemplateURL` is optional as far as the task is concerned -- isn't it? (i.e., notification without payload?). I agree we verify `url`. I also agree we can verify `payloadTemplateURL`s within the slack and GitHub task templates.

sriumcp (dismissed):
lgtm

charts/iter8/templates/_experiment.tpl
```diff
@@ -14,8 +14,12 @@ spec:
   {{- include "task.http" $.Values.http -}}
   {{- else if eq "ready" . }}
   {{- include "task.ready" $ -}}
+  {{- else if eq "slack" . }}
+  {{- include "task.slack" $.Values.slack -}}
+  {{- else if eq "github" . }}
+  {{- include "task.github" $.Values.github -}}
   {{- else }}
-  {{- fail "task name must be one of assess, custommetrics, grpc, http, or ready" -}}
+  {{- fail "task name must be one of assess, custommetrics, grpc, http, ready, github, or slack" -}}
```
kalantar, line 22: Not specific to this PR but this approach does not seem scalable to me. Perhaps we need a reference to a list of valid task names instead.

sriumcp (approved):
lgtm

Alan-Cha:
See #1322

PR #1324, opened 2022-08-25
Add notify templates

Alan-Cha:


charts/iter8/templates/_task-slack.tpl
```diff
@@ -0,0 +1,16 @@
...
+{{- end }}
+{{- if not .url }}
+{{- fail "please set a value for the url parameter" }}
+{{- end }}
+# task: send a Slack notification
+- task: notify
+  with:
+    url: {{ .url }}
+    method: POST
+    payloadTemplateURL: {{ default "https://github.com/iter8-tools/iter8/master/blob/slack-template.tpl" .payloadTemplateURL }}
```
sriumcp: Is this where the "default" slack template will be located for v0.11? I don't see it there. Perhaps it should be located under .../iter8/master/blob/templates/slack-template.tpl (and the tpl files should also be added)?

sriumcp: Actually, it makes sense to locate them here: https://github.com/iter8-tools/iter8/tree/master/charts/iter8/templates

sriumcp (changes requested):

charts/iter8/templates/_task-slack.tpl
Alan-Cha: Fixed

Alan-Cha: @sriumcp Should this be tied down to a particular commit/branch? Because once we move to v0.12, this will be gone right? And this reference will not work anymore. Then users using v0.11 will not be able to use these tasks.

sriumcp: You are correct. We want to tie it to a commit. So, perhaps, let us push the templates first, create a commit, and then update/merge this?

Alan-Cha: Okay sounds good

sriumcp (approved):
lgtm

PR #1335, opened 2022-09-15
Linting for the project

sriumcp:
This PR adds listing to Iter8 repo.

This is something that must have been done a long time ago (we do have `staticcheck`).

This PR adds what [this style guide](https://github.com/uber-go/guide/blob/master/style.md#lint-runners) considers to be the minimal amount of linting that any go project should do.

To locally lint your code, you can run:
```shell
brew install golangci-lint
brew upgrade golangci-lint
golangci-lint run
```


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,46 @@
+name: golangci-lint
```
sriumcp, line 1: This yaml file is taken from [here](https://github.com/golangci/golangci-lint-action).

abn/application/application_test.go
```diff
@@ -76,7 +76,7 @@ func TestWrite(t *testing.T) {
 	a.Tracks["foo"] = "v1"
 
 	// Write writes immediately
-	Applications.Write(a)
+	_ = Applications.Write(a)
```
sriumcp, line 79: Ignore errors in tests, if you do not use them in asserts.

abn/application/applicationmap.go
```diff
@@ -239,10 +239,18 @@ func (m *ThreadSafeApplicationMap) BatchedWrite(a *Application) error {
 	lastWrite, ok := m.lastWriteTimes[a.Name]
 	if !ok || lastWrite == nil {
 		// no record of the application ever being written; write it now
-		m.Write(a)
+		err := m.Write(a)
+		if err != nil {
```
sriumcp, line 244: Never ignore real world errors in actual code.

abn/application/test_helper.go
```diff
@@ -21,7 +21,7 @@ func readYamlFromFile(folder, file string) ([]byte, error) {
 	_, filename, _, _ := runtime.Caller(1) // one step up the call stack
 	fname := filepath.Join(filepath.Dir(filename), folder, file)
 
-	return ioutil.ReadFile(fname)
+	return os.ReadFile(fname)
```
sriumcp: ioutil is deprecated.

abn/core/lookup.go
```diff
@@ -67,7 +67,7 @@ func rendezvousGet(a *abnapp.Application, user string) string {
 // hash computes the score for a version, user combination
 func hash(version, user string) uint64 {
 	var hasher maphash.Hash
-	hasher.WriteString(user)
-	hasher.WriteString(version)
+	_, _ = hasher.WriteString(user)
```
sriumcp, line 70: WriteString is guaranteed to never fail. Ok to ignore error.

sriumcp (commented):
Self review.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.17
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.29
```
kalantar: Why are we using v1.29 here? When I install manually, I see version 1.49.0

kalantar:
The referenced style guide suggests using a specific set of linters and provides a config file to use to achieve this. We are not doing so; we are using the default set that comes with `golangci-lint` which is different (and not a superset).

**Recommended minimal set:**

- errcheck
- goimports
- golint
- govet
- staticcheck

**Set actually being used:**

- errcheck
- gosimple
- govet
- iineffassign
- staticcheck
- typecheck
- unused

There are a lot of linters but we need some reasonable / consistent decision here.

kalantar:
Can we add a make target for this as well? When I remember, I like to run as many of the checks (and tests) locally before committing changes.

abn/application/application_test.go
kalantar: I gotta be honest, I find  it annoying to be explicit about this every time.

abn/application/applicationmap.go
kalantar: So this is going to require some thinking. Errors are sometimes ignored for a reason. In this case, if `Write` fails, it fails. Agree that logging the error is probably a good idea. But I don't think we always need to return it.

Reviewed all uses of this method (`BatchedWrite`) and it is ok to return it.

abn/core/write_metric.go
```diff
@@ -39,7 +39,5 @@ func writeMetricInternal(application, user, metric, valueStr string) error {
 	m.Add(value)
 
 	// persist updated metric
-	abnapp.Applications.BatchedWrite(a)
-
-	return err
+	return abnapp.Applications.BatchedWrite(a)
```
kalantar, line 42: Line 35 should probably return the error instead of `nil` as well.

action/report/htmlreporter.go
```diff
@@ -53,57 +54,6 @@ func (ht *HTMLReporter) Gen(out io.Writer) error {
 	return nil
 }
 
-// HTMLHistCharts returns histogram charts section in HTML report
```
kalantar, line 56: I'm confused. Are these changes related to the lint report? If not, what is being attempted here?

base/collect_http.go
```diff
@@ -247,7 +250,9 @@ func (t *collectHTTPTask) run(exp *Experiment) error {
 			Description: "number of requests sent",
 			Type:        CounterMetricType,
 		}
-		in.updateMetric(m, mm, 0, float64(data.DurationHistogram.Count))
+		if err = in.updateMetric(m, mm, 0, float64(data.DurationHistogram.Count)); err != nil {
```
kalantar, line 252: I guess my question is: do we want to fail immediately if we can't update one metric of the set?  Or do we wish to continue in a best effort way?  Applies to all `updateMetric` calls

cmd/abn.go
```diff
@@ -29,8 +29,3 @@ func newAbnCmd() *cobra.Command {
 	}
 	return cmd
 }
-
-// initialize with assert
-func init() {
-	rootCmd.AddCommand(newAbnCmd())
-}
```
kalantar, line 36: Please explain this change. Why?

kalantar (changes requested):

.github/workflows/golangci-lint.yml
sriumcp: Fixed.

abn/application/application_test.go
sriumcp: Me too. But when it comes to code quality, as well as security, consistency is key. The linters make it reasonably easy to write high quality code in line with community standards. Of course, once we learn more about these linters, we can revisit how to do more fine grained configuration of these linters based on our tastes. For now, going with their suggestion seems very reasonable.

abn/application/applicationmap.go
sriumcp: This is exactly right. Whenever you ignore errors, do so explicitly, and not silently.

abn/core/write_metric.go
sriumcp: Done.

sriumcp (commented):
Self review.

action/report/htmlreporter.go
sriumcp: Which changes? HTMLHistCharts was an unused func which is now gone.

sriumcp:
> Can we add a make target for this as well? When I remember, I like to run as many of the checks (and tests) locally before committing changes.

Done.

cmd/abn.go
sriumcp: The order of `init()` executions is arbitrary in a golang program. It turns out that some of the init funcs for child flags (especially, the init func for the child flag `krun`) was assuming stuff that was supposed to happen in the init func of the parent flags (`k`).

I was able to catch this by **not** ignoring errors, and reporting them, which caused my tests to fail, which in turn enabled me to fix the situation.

So, a combination of linting + testing + proper error handing ... leading to correct code ... is what is at work here.

base/collect_http.go
sriumcp: This is a good point. For now, let us fail immediately (IMO, failing immediately, rather than ignoring silently is the improvement here). We can get a bit more experience with custom metrics, and then revisit this issue.

sriumcp: Note that `collect_http` and `collect_grpc` have predefined metric names, which means update metric really should not fail, which means (if for some unknown reason it does), we really want to catch and return the error (so yes, failing immediately makes sense in these files).

action/report/htmlreporter.go
kalantar: unused. ah. ok. didn't realize; thought perhaps there were some report changes as well.

kalantar (approved):

PR #1338, opened 2022-09-19
read abn task

kalantar:

### DIFF
--- a/abn/application/application.go
+++ b/abn/application/application.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/iter8-tools/iter8/base/log"
+	"github.com/iter8-tools/iter8/base/metrics"
 )
 
 // Application is an application observed in a kubernetes cluster
@@ -50,7 +51,7 @@ func (a *Application) GetVersion(version string, allowNew bool) (*Version, bool)
 		if allowNew {
 			log.Logger.Debugf("GetVersion no data found; returning %+v", v)
 			v = &Version{
-				Metrics: map[string]*SummaryMetric{},
+				Metrics: map[string]*metrics.SummaryMetric{},
 			}
 			a.Versions[version] = v
 			return v, true
@@ -79,7 +80,7 @@ func (a *Application) UnmarshalJSON(data []byte) error {
 	}
 	for _, v := range a.Versions {
 		if v.Metrics == nil {
-			v.Metrics = map[string]*SummaryMetric{}
+			v.Metrics = map[string]*metrics.SummaryMetric{}
 		}
 	}
 
--- a/abn/application/application.go
+++ b/abn/application/application.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	"github.com/iter8-tools/iter8/base/log"
-	"github.com/iter8-tools/iter8/base/metrics"
+	"github.com/iter8-tools/iter8/base/summarymetrics"
 )
 
 // Application is an application observed in a kubernetes cluster
@@ -51,7 +51,7 @@ func (a *Application) GetVersion(version string, allowNew bool) (*Version, bool)
 		if allowNew {
 			log.Logger.Debugf("GetVersion no data found; returning %+v", v)
 			v = &Version{
-				Metrics: map[string]*metrics.SummaryMetric{},
+				Metrics: map[string]*summarymetrics.SummaryMetric{},
 			}
 			a.Versions[version] = v
 			return v, true
@@ -80,7 +80,7 @@ func (a *Application) UnmarshalJSON(data []byte) error {
 	}
 	for _, v := range a.Versions {
 		if v.Metrics == nil {
-			v.Metrics = map[string]*metrics.SummaryMetric{}
+			v.Metrics = map[string]*summarymetrics.SummaryMetric{}
 		}
 	}
 
--- a/abn/core/service.go
+++ b/abn/core/service.go
@@ -9,8 +9,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"os/signal"
-	"syscall"
 
 	pb "github.com/iter8-tools/iter8/abn/grpc"
 	"github.com/iter8-tools/iter8/abn/k8sclient"
@@ -20,7 +18,6 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/emptypb"
 
-	"helm.sh/helm/v3/pkg/cli"
 	// auth package is necessary to enable authentication with various cloud vendors
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 )
@@ -36,26 +33,7 @@ var (
 )
 
 // Start is entry point to configure services and start them
-func Start() {
-	w := initializeServer()
-	stopCh := make(chan struct{})
-
-	// start watchers
-	go w.Start(stopCh)
-
-	// launch gRPC server to respond to frontend requests
-	go launchGRPCServer([]grpc.ServerOption{})
-
-	// shutdown gracefully on termination
-	sigCh := make(chan os.Signal, 1)
-	signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
-
-	<-sigCh
-	close(stopCh)
-}
-
-func initializeServer() *watcher.Iter8Watcher {
-	k8sclient.Client = *k8sclient.NewKubeClient(cli.New())
+func Start(stopCh chan struct{}) error {
 	if err := k8sclient.Client.Initialize(); err != nil {
 		log.Logger.WithStackTrace("unable to initialize k8s client").Fatal(err)
 	}
@@ -69,7 +47,14 @@ func initializeServer() *watcher.Iter8Watcher {
 	// set up resource watching as defined by config
 	c := readConfig(abnConfigFile)
 	w := watcher.NewIter8Watcher(c.Resources, c.Namespaces)
-	return w
+
+	// start watchers
+	go w.Start(stopCh)
+
+	// launch gRPC server to respond to frontend requests
+	go launchGRPCServer([]grpc.ServerOption{})
+
+	return nil
 }
 
 // newServer returns a new gRPC server
--- a/abn/core/service.go
+++ b/abn/core/service.go
@@ -27,11 +27,6 @@ const (
 	watcherConfigEnv = "WATCHER_CONFIG"
 )
 
-// var (
-// 	// port the service listens on
-// 	port = flag.Int("port", 50051, "The server port")
-// )
-
 // Start is entry point to configure services and start them
 func Start(port int, stopCh chan struct{}) error {
 	if err := k8sclient.Client.Initialize(); err != nil {
@@ -107,6 +102,8 @@ func (server *abnServer) GetApplicationData(ctx context.Context, metricReqMsg *p
 
 // launchGRPCServer starts gRPC server
 func launchGRPCServer(port int, opts []grpc.ServerOption, stopCh chan struct{}) {
+	log.Logger.Tracef("starting gRPC service on port %d", port)
+
 	lis, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", port))
 	if err != nil {
 		log.Logger.WithError(err).Fatal("failed to listen")
--- a/abn/core/service.go
+++ b/abn/core/service.go
@@ -5,10 +5,10 @@ package core
 
 import (
 	"context"
-	"flag"
 	"fmt"
 	"net"
 	"os"
+	"sync"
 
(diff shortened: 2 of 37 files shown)


kalantar:
Output of (text) report now shows track names if known:

```sh
$ iter8 k launch -l trace --noDownload --set abnmetrics.application=default/backend --set "tasks={abnmetrics,assess}" --set runner=cronjob --set cronjobSchedule="*/1 * * * *"
```
```
Experiment summary:
*******************

  Experiment completed: true
  No task failures: true
  Total number of tasks: 1
  Number of completed tasks: 1

Latest observed values for metrics:
***********************************

  Metric                   | candidate | default
  -------                  | -----     | -----
  abn/sample_metric/count  | 765.00    | 733.00
  abn/sample_metric/max    | 100.00    | 100.00
  abn/sample_metric/mean   | 50.11     | 49.64
  abn/sample_metric/min    | 0.00      | 0.00
  abn/sample_metric/stddev | 28.63     | 29.25
```

sriumcp:
@kalantar Could you also provide a sample html report output?

abn/core/service.go
```diff
@@ -37,6 +37,24 @@ var (
...
+
+	// launch gRPC server to respond to frontend requests
+	go launchGRPCServer([]grpc.ServerOption{})
+
+	// shutdown gracefully on termination
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
+
+	<-sigCh
+	close(stopCh)
```
sriumcp: The  `stopCh` helps to gracefully shut down the informers. 

Similarly, we might also want to add a few lines in `launchGRPCServer` func to gracefully shut down the grpc server.

[Here](https://gist.github.com/embano1/e0bf49d24f1cdd07cffad93097c04f0a) is an example ...

```go
	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", 8080))
	if err != nil {
		log.Fatalf("failed to listen: %v", err)
	}
	grpc := grpc.NewServer()
	pb.RegisterGreeterServer(grpc, &gSrv)

	sigCh := make(chan os.Signal, 1)
	signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
	wg := sync.WaitGroup{}
	wg.Add(1)
	go func() {
		s := <-sigCh
		log.Printf("got signal %v, attempting graceful shutdown", s)
		cancel()
		grpc.GracefulStop()
		// grpc.Stop() // leads to error while receiving stream response: rpc error: code = Unavailable desc = transport is closing
		wg.Done()
	}()

	log.Println("starting grpc server")
	err = grpc.Serve(lis)
	if err != nil {
		log.Fatalf("could not serve: %v", err)
	}
	wg.Wait()
	log.Println("clean shutdown")
```

abn/core/service.go
```diff
@@ -37,6 +37,24 @@ var (
 
 // Start is entry point to configure services and start them
 func Start() {
+	w := initializeServer()
```
sriumcp: In `autox`, we structure the Start function like [this](https://github.com/iter8-tools/iter8/blob/3038ba07e857966209535b84f5736c33a39e74d5/autox/watcher.go#L20). 

This implies, we structure the `cmd` like [this](https://github.com/iter8-tools/iter8/blob/3038ba07e857966209535b84f5736c33a39e74d5/cmd/autox.go#L26)

More importantly, we were able to test it like [this](https://github.com/iter8-tools/iter8/blob/3038ba07e857966209535b84f5736c33a39e74d5/autox/watcher_test.go#L31) -- the key point here is how the test simply calls `Start(...)` and does not copy and paste the internal code from  the `Start(...)` func inside the test func.

sriumcp: Please structure the `abn` Start(...) similar to the `autox` Start so that it can be tested the same way (without copying the internals of Start into the test func).

abn/core/service_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"path/filepath"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc"
+)
+
+func TestService(t *testing.T) {
```
sriumcp: This is a good test for `initializeServer` func. This is not quite the test for the Service itself... please see comment above on directly calling `Start` (instead of copying and pasting the internals of Start into the test func).

action/report/textreporter.go
```diff
@@ -81,19 +81,32 @@ func (tr *TextReporter) getSLOStrText(i int, upper bool) (string, error) {
 	return str, nil
 }
 
+func (tr *TextReporter) printVersions(w *tabwriter.Writer) {
```
sriumcp, line 84: Please incorporate similar logic for HTML reporter (for example, using a new method defined on the reporter).

base/abn_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+type ABNMetricsInputs struct {
+	// Application is name of application to evaluate
+	Application string `json:"application" yaml:"application"`
+}
+
+// collectABNMetricsTask is task defintion for abnmetrics task
+type collectABNMetricsTask struct {
+	TaskMeta
+	With   ABNMetricsInputs `json:"with" yaml:"with"`
+	client abnClientInterface
```
sriumcp, line 71: Fantastic pattern that enables mocking the interface in tests.

base/abn_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// initializeDefaults sets default values for the task
+func (t *collectABNMetricsTask) initializeDefaults() {
+	k8sclient.Client = *k8sclient.NewKubeClient(cli.New())
```
sriumcp: Why does this task need to initialize a k8sclient? For instance, the `collect_http` task which also writes metrics, doesn't appear to require the client initialization.

base/abn_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := k8sclient.Client.Initialize(); err != nil {
+		log.Logger.WithStackTrace("unable to initialize k8s client").Fatal(err)
+	}
+
+	if t.client == nil {
+		t.client = &defaultABNClient{
+			endpoint: "iter8-abn:50051",
```
sriumcp: Please make this endpoint a parameter to the task. We can set default values in Helm instead of hardcoding them here.

base/experiment.go
```diff
@@ -76,6 +78,9 @@ type Insights struct {
 	// NumVersions is the number of app versions detected by Iter8
 	NumVersions int `json:"numVersions" yaml:"numVersions"`
 
+	// VersionNames is list of version identifiers if known
```
sriumcp, line 81:
```go
VersionNames []VersionInfo

type VersionInfo struct {
  Version string
  Track  string
}
```

The above is the preferred data structure. This way you would be able to output names like this -- `baseline (x9183)` and `candidate (y2849)`. The things in the parentheses are the version names, which might often be unintelligible even to the developer/ops person (because the version names might be based off of commit IDs, COS buckets, etc.).

base/metrics/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package metrics provides a summary metric implementation
```
sriumcp: Please rename to `package summarymetrics`.

sriumcp (changes requested):
Initial comments.

abn/application/application.go
```diff
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/iter8-tools/iter8/base/log"
+	"github.com/iter8-tools/iter8/base/metrics"
```
sriumcp: `summarymetrics` package name

base/abn_metrics.go
```diff
@@ -0,0 +1,166 @@
...
+	// initialize insights in Result with number of tracks
+	err = exp.Result.initInsightsWithNumVersions(numTracks)
+	if err != nil {
+		return err
+	}
+	log.Logger.Tracef("intialized insights with %d versions", numTracks)
+
+	in := exp.Result.Insights
+
+	// add metrics for tracks
```
sriumcp: More comments on what is going in the for loop would be helpful in review.

kalantar:
> @kalantar Could you also provide a sample html report output?

image

base/experiment.go
kalantar: done

base/metrics/doc.go
kalantar: done

base/abn_metrics.go
kalantar: no longer needed. good catch. removed.

action/report/textreporter.go
kalantar: done

abn/core/service.go
kalantar: much nicer pattern. thx.

sriumcp (changes requested):

abn/core/service_test.go
kalantar: modified along with structure of `Start()`

abn/core/service.go
kalantar: done

base/abn_metrics.go
kalantar: done. now supports option `--set abnmetrics.endpoint=iter8-abn:50051`. Defaults to this value.

base/abn_metrics.go
kalantar: done

abn/application/application.go
kalantar: done

base/abn_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+type collectABNMetricsTask struct {
+	TaskMeta
+	With   ABNMetricsInputs `json:"with" yaml:"with"`
+	client abnClientInterface
+}
+
+// initializeDefaults sets default values for the task
+func (t *collectABNMetricsTask) initializeDefaults() {
+	if t.client == nil {
+		ep := defaultEndpoint
```
sriumcp: Could we please 

1. Eliminate `defaultEndpoint`
2. Make `t.With.Endpoint` required (i.e., add it to validateInputs, and make sure it exists)
3. Set:
```
t.client = &defaultABNClient{
  endpoint: t.With.Endpoint
}
```

Basically, we do not want to deal with this setting of this end point within this task at all (but simply require that whoever invokes the task pass it along).

Especially, since we said Helm chart would set a default value (we don't want two defaults, or two different places where they are managed).

base/abn_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+	// for each track (and corresponding version), get the Version object
+	// Use it to update all metrics for this version
+	for track, version := range a.Tracks {
+		// set the track identifier/version name in result
+		in.VersionNames[versionIndex].Version = version
+		in.VersionNames[versionIndex].Track = track
+		// get version object from retrieved application object
+		v, _ := a.GetVersion(version, false)
+		if v == nil {
+			log.Logger.Debugf("expected version %s not found", version)
```
sriumcp: The usual pattern we have used here is:
```go
err := fmt.Errorf("expected version %s not found", version)
log.Logger.Error(err)
return err
```

Simply makes the printed error and returned error consistent.

***

Should the log level be Warn instead of Debug? Not sure...

base/abn_metrics.go
```diff
@@ -0,0 +1,174 @@
...
+			return errors.New("expected version not found")
+		}
+		log.Logger.Tracef("version %s is mapped to track %s; using index %d", version, track, versionIndex)
+		// update all metrics with new values (is summary metric so just replace)
+		for metric, m := range v.Metrics {
+			log.Logger.Tracef("   updating metric %s with data %+v", metric, m)
+			err := in.updateMetric(
+				abnMetricProvider+"/"+metric,
+				MetricMeta{
+					Description: "summary metric",
```
sriumcp: The description and units are problematic... I now see that. We don't want to complicate the SDK through these fields ... but in the future, it would be useful to figure out how the user may supply these for better reporting.

sriumcp (changes requested):
Few more changes requested.

base/abn_metrics.go
sriumcp: For now, the above looks good.

kalantar: Description is not necessary and it isn't clear that this description adds much value. I'll remove it.

base/abn_metrics.go
kalantar: done

base/abn_metrics.go
kalantar: done.

sriumcp (dismissed):

sriumcp (approved):
lgtm

Issue #1334, opened 2022-09-14
Ensure uniqueness of chartGroup names in autox config


sriumcp:
**Is your feature request related to a problem? Please describe.**
The current way of specifying the chart group could lead to naming conflicts. Specifically, multiple groups could end up with the same name.

**Solution:**
```yaml
chartGroupName1:
  charts:
    releaseSpec1:
      helmRepo: abc
      chartName: def
      valuesFileUrl: abcdef.com/abcdef
    releaseSpec2:
      helmRepo: ghi
      chartName: jkl
      valuesFileUrl: ghijkl.com/ghijkl
chartGroupName2:
  charts:
    releaseSpec3: 
      helmRepo: mno
      chartName: pqr
      valuesFileUrl: mnopqr.com/mnopqr
```

sriumcp:
Assume autox is watching for deployment objects in the test namespace.

Suppose the user does the following.

```shell
kubectl create deployment nginx --image=nginx -n test
kubectl label deploy/nginx iter8.tools/autox-group=myApp -n test
```

AutoX needs to create a couple of Helm releases, corresponding to the two charts in this chart group. The name of a release will be `autox-<chart-group-name>-<chart-name>`.

So, `autox-myApp-name1`, `autox-myApp-name2`, etc., are the releases we will get.

sriumcp:
During an `addObject` event -- assuming the object has appropriate Iter8 labels, if there already exist corresponding releases, we upgrade them. If not we create them.

During an `updateObject` event; check if labels changed -- if not, nothing to do; else, treat it as a deletion event (deletion of the old object) followed by an addition event (addition of the new object).

During an `deleteObject` event; check if the deleted object has iter8 labels -- if there is an Iter8 label, delete the corresponding release.

sriumcp:
Concurrency: none.

Ensure that all handler invocations run in sequence through a mutex.

Example...
```go
mu.Lock()
defer mu.Unlock()
// invoke handler here ... 
}
```

Here `mu` is a sync.Mutex object.

PR #1339, opened 2022-09-19
Reformat chart group config

Alan-Cha:


autox/config.go
```diff
@@ -34,12 +34,10 @@ type chart struct {
 }
 
 type chartGroup struct {
-	Name string `yaml:"name"`
-
-	Charts []chart `yaml:"charts"`
+	Charts map[string]chart `yaml:"charts"`
```
sriumcp: Add `json` tags wherever you have `yaml` tags.

sriumcp: Also, please refer to #1334 for renaming of fields.

autox/informer.go
```diff
@@ -3,29 +3,119 @@ package autox
...
+	// TODO: check if there is a preexisting Helm release
+
+	// TODO: delete Helm chart
+
+	log.Logger.Debug("Delete chart:", releaseName)
 }
 
-var deleteObject = func(obj interface{}) {
-	log.Logger.Debug("Delete:", obj)
+func iterateCharts(appName string, cgc chartGroupConfig, callback func(releaseName string, chart chart)) {
```
sriumcp: `doChartAction` instead of `iterateCharts`.

sriumcp: golang enum typed as `chartAction` that can be release or delete. Enums are described [here](https://www.sohamkamani.com/golang/enums/#enum-types).

```
func doChartAction(appName string, cgc chartGroupConfig, act chartAction)
```

autox/informer.go
```diff
@@ -3,29 +3,119 @@ package autox
...
+	autoxLabel = "iter8.tools/autox-group"
+)
+
+func getReleaseName(chartGroupName string, chartName string) string {
+	return fmt.Sprintf("autox-%s-%s", chartGroupName, chartName)
 }
 
-var updateObject = func(oldObj, obj interface{}) {
-	log.Logger.Debug("Update:", oldObj, obj)
+func releaseHelmChart(releaseName string, chart chart) {
```
sriumcp: `releaseAction` or `deleteAction`

sriumcp (changes requested):
comments we discussed during the review.

autox/informer.go
```diff
@@ -3,36 +3,144 @@ package autox
...
+
+	uOldObj := oldObj.(*unstructured.Unstructured)
+	oldLabels := pruneLabels(uOldObj.GetLabels())
+
+	uObj := obj.(*unstructured.Unstructured)
+	resourceName := uObj.GetName()
+	// example label: iter8.tools/autox-group=hello
+	labels := pruneLabels(uObj.GetLabels())
+
+	// if reflect.DeepEqual(oldLabels, labels) { return }
```
Alan-Cha: I was not sure how to get the difference of maps. Any suggestions, @sriumcp?

sriumcp: We actually don't need difference of maps at all.

```go
if reflect.DeepEqual(oldLabels, labels) { 
  return 
}
// the two lines below will only kick in when the label sets are unequal.
deleteHelmReleases(releaseSpecName, releaseSpec, oldLabels) // implementation stubbed for now
installHelmReleases(releaseSpecName, releaseSpec, labels) // implementation stubbed for now
```

autox/config.go
```diff
@@ -15,31 +15,29 @@ import (
...
 
 type chartGroup struct {
-	Name string `yaml:"name"`
-
-	Charts []chart `yaml:"charts"`
+	Charts map[string]chart `json:"charts" yaml:"charts"`
 }
 
-type chartGroupConfig []chartGroup
+type chartGroupConfig map[string]chartGroup
```
kalantar, line 44: I would find it helpful to have comments on types explaining for example what the key is.

autox/informer.go
```diff
@@ -3,36 +3,146 @@ package autox
...
 var addObject = func(obj interface{}) {
 	log.Logger.Debug("Add:", obj)
+
+	uObj := obj.(*unstructured.Unstructured)
+	resourceName := uObj.GetName()
+	// example label: iter8.tools/autox-group=hello
+	labels := uObj.GetLabels()
+
+	// check if the app name matches the name in the autox label
+	if autoxLabelName, ok := labels[autoxLabel]; ok && resourceName == autoxLabelName {
```
kalantar, line 86: I'm confused by why add and delete use this approah and update uses prunelables and compare. I would have thought it would be the same

autox/config.go
Alan-Cha: Done

autox/informer.go
Alan-Cha: You are right. I should have also updated the app and delete functions as well.

sriumcp (approved):
lgtm

PR #1344, opened 2022-09-29
Chaos readiness check

sriumcp:
There are no conditions in the chaos CR. Hence, the check is only for existence.


sriumcp:
Replaced by #1345

PR #1345, opened 2022-09-29
check existence of chaosengines resource object

sriumcp:


kalantar (changes requested):
I believe you need to modify `_k-role.tpl` as well to git Iter8 permission to get the resource.

sriumcp:
> I believe you need to modify `_k-role.tpl` as well to git Iter8 permission to get the resource.

Absolutely. Thanks for catching. Fixed now.

kalantar (approved):

Alan-Cha (approved):

Issue #1352, opened 2022-10-13
Move iter8 chart to iter8-tools/hub repository


kalantar:
Today the `iter8` command uses a single `iter8` Helm chart to create experiment definitions. The `iter8` chart is defined in the `iter8-tools/iter8` repository. 

The implementation downloads the chart from the file system assuming it is a GitHub repository. However, it is structured as a Helm chart and treating it as such would simplify the implementation of Iter8. To do this:

1. move the chart to the `iter8-tools/hub` repository; the hub repository is already exported as a Helm repository 
2. modify `iter8` commands to use this Helm repository.

In the process, we want to be sure not to disrupt the development of automated experiments (i.e., `autox`) and we want to continue to support execution using a local chart for chart development use cases.

kalantar:
The current implementation of `iter8 launch` and `iter8 k launch` downloads the chart (assuming it is from a GitHub repository) using the internal implementation of `iter8 hub`), processes the downloaded template using the internal implementation of `iter8 gen` and runs the resulting experiment (file or secret) using the internal implementation of `iter8 run`. 

While something similar is still needed for `iter8 launch`, a more Helm like implementation of `iter8 k launch` could more easily be accomplished using the equivalent of `helm install` and/or `helm upgrade`.

By default, the Helm repository and chart name are defaulted (to `https://iter8-tools.github.io/hub` and `iter8`, respectively). The launch command is then similar to:
```shell
iter8 --set "tasks={http}" --set http.url=http://httpbin.default/get --set runner=job
```

Optionally, the Helm repository and chart can set using options:
- `--repo <helm_repository_url>`  identifies a Helm repository by URL
- `--chartName <chart_name>` identifies a chart name

Furthermore, Helm allows identifying a chart using the form `repo/chart` where `repo` is a name added using the `helm repo add` command.

In all cases where a remote repository is specified, Helm commands offer a set of access options (`--ca-file`, `--cert-file`, `--key-file`, `--insecure-skip-verify`, `--keyring`, `--password`, `--pass-credentials`, `--username` and `--verify`

Finally, to indicate that a chart is local, `--localChart` can be specified.

When charts are pulled using the Helm APIs, there is no need to update the dependencies; the latest dependencies are already pulled.

The same options can be made available to local execution (`iter8 launch`).

In summary, the following changes are to be made to the commands
- `iter8 gen` - remove update dependencies; hide
- `iter8 hub` - remove this command (user can use `helm pull` if desired instead)
[...]

PR #1353, opened 2022-10-13
github repo -> helm repo

kalantar:
Addresses: https://github.com/iter8-tools/iter8/issues/1352


kalantar:
This PR depends on https://github.com/iter8-tools/hub/pull/31.
It represents a breaking change.
Once the above PR is merged, `testdata/charts/iter8` should be updated using:

```shell
helm pull --repo https://iter8-tools.github.io/hub iter8 --untar
```

cmd/launch.go
```diff
@@ -74,8 +71,35 @@ func addDryRunFlag(cmd *cobra.Command, dryRunPtr *bool) {
...
+func addLocalChartFlag(cmd *cobra.Command, localChartPtr *bool) {
+	cmd.Flags().BoolVar(localChartPtr, "localChart", false, "use local charts")
+	cmd.Flags().Lookup("localChart").NoOptDefVal = "true"
+}
+
+// addChartPathOptionsFlags adds flags related to Helm chart repository
+// copied from
+// https://github.com/helm/helm/blob/ce66412a723e4d89555dc67217607c6579ffcb21/cmd/helm/flags.go
+func addChartPathOptionsFlags(cmd *cobra.Command, c *action.ChartPathOptions) {
+	cmd.Flags().StringVar(&c.Version, "version", "", "specify a version constraint for the chart version to use. This constraint can be a specific tag (e.g. 1.1.1) or it may reference a valid range (e.g. ^2.0.0). If this is not specified, the latest version is used")
```
sriumcp: User `majorMinor` version to derive version automatically (e.g., `0.12.x`). Comment out other options.

driver/kubedriver.go
```diff
@@ -300,14 +271,32 @@ func writeManifest(rel *release.Release) error {
...
+	client.ChartPathOptions.CertFile = chartPathOptions.CertFile
+	client.ChartPathOptions.KeyFile = chartPathOptions.KeyFile
+	client.ChartPathOptions.InsecureSkipTLSverify = chartPathOptions.InsecureSkipTLSverify
+	client.ChartPathOptions.Keyring = chartPathOptions.Keyring
+	client.ChartPathOptions.Password = chartPathOptions.Password
+	client.ChartPathOptions.PassCredentialsAll = chartPathOptions.PassCredentialsAll
+	client.ChartPathOptions.RepoURL = chartPathOptions.RepoURL
+	client.ChartPathOptions.Username = chartPathOptions.Username
+	client.ChartPathOptions.Verify = chartPathOptions.Verify
+	client.ChartPathOptions.Version = chartPathOptions.Version
```
sriumcp, line 290: Version is the only property that need to be derived.

driver/kubedriver.go
```diff
@@ -300,14 +271,32 @@ func writeManifest(rel *release.Release) error {
...
 
-	ch, vals, err := kd.getChartAndVals(chartDir, valueOpts)
+	client.ChartPathOptions.CaFile = chartPathOptions.CaFile
+	client.ChartPathOptions.CertFile = chartPathOptions.CertFile
+	client.ChartPathOptions.KeyFile = chartPathOptions.KeyFile
+	client.ChartPathOptions.InsecureSkipTLSverify = chartPathOptions.InsecureSkipTLSverify
+	client.ChartPathOptions.Keyring = chartPathOptions.Keyring
+	client.ChartPathOptions.Password = chartPathOptions.Password
+	client.ChartPathOptions.PassCredentialsAll = chartPathOptions.PassCredentialsAll
+	client.ChartPathOptions.RepoURL = chartPathOptions.RepoURL
```
sriumcp, line 287: RepoURL is also derived automatically (fixed for now).

driver/kubedriver.go
```diff
@@ -323,46 +312,53 @@ func (kd *KubeDriver) upgrade(chartDir string, valueOpts values.Options, group s
...
+
+	// buf := new(bytes.Buffer)
 	client := action.NewInstall(kd.Configuration)
 	client.Namespace = kd.Namespace()
+	client.RepoURL = chartPathOptions.RepoURL
 	client.DryRun = dry
 	client.ReleaseName = group
 
-	ch, vals, err := kd.getChartAndVals(chartDir, valueOpts)
+	client.ChartPathOptions.CaFile = chartPathOptions.CaFile
```
sriumcp: Comment out properties we are not using for now.

sriumcp (changes requested):
Please see requested changes.

cmd/launch.go
kalantar: Set defaults in `action.NewLaunchOpts()` for repoURL and version. Commented out `addChartPathOptionsFlags()`. This removes all CLI options to set the chart path options (including `--repo`).

driver/kubedriver.go
kalantar: Left this as is; it is merely copying the values to input for the helm command. These are either explicitly set earlier or are default values.

driver/kubedriver.go
kalantar: Left this as is; it is merely copying the values to input for the helm command. These are either explicitly set earlier or are default values.

driver/kubedriver.go
kalantar: see other comments

driver/kubedriver.go
```diff
@@ -323,46 +313,54 @@ func (kd *KubeDriver) upgrade(chartDir string, valueOpts values.Options, group s
...
+	// buf := new(bytes.Buffer)
 	client := action.NewInstall(kd.Configuration)
 	client.Namespace = kd.Namespace()
+	client.RepoURL = chartPathOptions.RepoURL
 	client.DryRun = dry
 	client.ReleaseName = group
 
-	ch, vals, err := kd.getChartAndVals(chartDir, valueOpts)
+	// copy chartPathOptions to client
+	client.ChartPathOptions.CaFile = chartPathOptions.CaFile
```
sriumcp: Could we please remove all of this? We don't even plan to document how the user is intended to use any of these options ... for now, we are better of without this.

We do need to discuss binary/image/chart signing and verification. We will revisit all of these options at that time.

driver/kubedriver.go
```diff
@@ -300,14 +271,33 @@ func writeManifest(rel *release.Release) error {
...
 // Upgrade a Kubernetes experiment to the next release
-func (kd *KubeDriver) upgrade(chartDir string, valueOpts values.Options, group string, dry bool) error {
+func (kd *KubeDriver) upgrade(chartPathOptions action.ChartPathOptions, chartName string, valueOpts values.Options, group string, dry bool) error {
 	client := action.NewUpgrade(kd.Configuration)
 	client.Namespace = kd.Namespace()
 	client.DryRun = dry
 
-	ch, vals, err := kd.getChartAndVals(chartDir, valueOpts)
+	// copy chartpathoptions
+	client.ChartPathOptions.CaFile = chartPathOptions.CaFile
```
sriumcp: Could we please remove all of this? We don't even plan to document how the user is intended to use any of these options ... for now, we are better of without this.

We do need to discuss binary/image/chart signing and verification. We will revisit all of these options at that time.

sriumcp (changes requested):
Please see change requests.

driver/kubedriver.go
kalantar: This code is initializing a field (the `ChartPathOptions` field) in the Helm object (`client`). This object is used both to find the chart and to call `RunWithContext()` to do the actual work. We do want to initialize the field. However, we may not need to copy every field. I can comment out the fields that we do not explicitly set.

driver/kubedriver.go
kalantar: same response

sriumcp (approved):
lgtm

PR #1355, opened 2022-10-14
Autoxhelm create ArgoCD application

Alan-Cha:


autox/application.tpl
```diff
@@ -0,0 +1,27 @@
+kind: Application
+metadata:
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+  ownerReferences:
+    - apiVersion: {{ .owner.apiVersion }}
```
sriumcp: Indent like in line 27.

sriumcp: Secret has a fixed apiVersion ("v1".. or something like that).

autox/application.tpl
```diff
@@ -0,0 +1,27 @@
...
+    kind: Secret
+    name: {{ .owner.name}}
+    uid: {{ .owner.uid}}
+spec:
+  destination:
+    namespace: kubeseal
```
sriumcp: For now, let us make this `{{ .trigger.Namespace }}`.

autox/application.tpl
```diff
@@ -0,0 +1,27 @@
...
+  source:
+    chart: {{ .chart.name }}
+    helm:
+      values: 
+        {{ .chart.values | toYAML | indent 4 }}
+    repoURL: {{ .chart.url }}
+    targetRevision: {{ .chart.version }}
+  syncPolicy:
+    automated:
+      prune: true
```
sriumcp, line 25: Document what `prune` and `selfHeal` mean.

autox/config.go
```diff
@@ -7,70 +7,81 @@ import (
...
+
+// 	// Name is the name of the Helm chart
+// 	Name string `json:"name" yaml:"name"`
+
+// 	// ValuesFileURL is the URL to the values file of the Helm chart
+// 	ValuesFileURL string `json:"valuesFileURL" yaml:"valuesFileURL"`
+// }
+
 // chart points to a particular Helm chart
 type chart struct {
```
sriumcp: This is actually a releaseSpec and not  a chart. So perhaps, rename `chart` to `releaseSpec`?

autox/config.go
```diff
@@ -7,70 +7,81 @@ import (
...
-// chartGroupConfig is the configuration for all the Helm charts
-// the key is the experiment group name
-type chartGroupConfig map[string]chartGroup
-
-// readResourceConfig reads yaml config file and converts to a resourceConfig object
-func readResourceConfig(fp string) (config resourceConfig) {
-	// empty configuration
-	config = resourceConfig{}
+// chartGroup is the configuration of all the Helm charts for a particular experiment group and their install trigger
+type chartGroup struct {
```
sriumcp: `releaseGroupSpec` instead of `chartGroup`.

autox/config.go
```diff
@@ -7,70 +7,81 @@ import (
...
-		return config // empty configuration
-	}
+	// ReleaseSpecs is the set of Helm charts
+	// the keys in ReleaseSpecs are identifiers for each chart
+	ReleaseSpecs map[string]chart `json:"releaseSpecs" yaml:"releaseSpecs"`
+}
 
-	log.Logger.Debugf("read configuration\n%s", string(yfile))
+// // chartGroupConfig is the configuration for all the Helm charts and their triggers
+type chartGroupConfig struct {
```
sriumcp: Remove `Namespaces`. 

Rename `chartGroupConfig` to `config`.

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	google.golang.org/api v0.62.0 // indirect
 	google.golang.org/grpc v1.45.0
 	google.golang.org/protobuf v1.28.0
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
sriumcp: Let us not use `gopkg.in/yaml.v2 v2.4.0`.

autox/informer.go
```diff
@@ -3,146 +3,364 @@ package autox
...
+		return err
+	} else if secretsLen > 1 {
+		log.Logger.Error("expected secret with label selector:", labelSelector, "but more than one were found")
+		return err
+	}
+	secret := secretList.Items[0]
+
+	values := applicationValues{
+		name:      releaseName,
+		namespace: namespace,
```
Alan-Cha, line 129: Trigger namespace?

autox/informer.go
```diff
@@ -3,146 +3,364 @@ package autox
...
+
+		owner: struct {
+			apiVersion string
+			name       string
+			uid        string
+		}{
+			apiVersion: secret.APIVersion,
```
Alan-Cha: Hard code `apiVersion`

autox/informer.go
```diff
@@ -3,146 +3,364 @@ package autox
...
+		log.Logger.Error("could not serialize application manifest")
+		return err
+	}
+	unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		log.Logger.Error("could not convert application manifest into unstructured object")
+		return err
+	}
+	unstructuredObj := &unstructured.Unstructured{Object: unstructuredMap}
+
```
Alan-Cha, line 156:
```
	a := &Application{}
	err = yaml.Unmarshal(rawData, a)
```

autox/k8sclient.go
```diff
@@ -68,10 +63,6 @@ func (c *kubeClient) init() (err error) {
 	return nil
 }
 
-// func (c *kubeClient) typed() kubernetes.Interface {
-// 	return c.clientset
-// }
-
 func (c *kubeClient) dynamic() dynamic.Interface {
```
sriumcp, line 69: This method looks like fluff... perhaps should be removed?

autox/k8sclient.go
```diff
@@ -30,7 +24,8 @@ type kubeClient struct {
 	dynamicClient dynamic.Interface
 }
 
-func newKubeClient(s *cli.EnvSettings) *kubeClient {
+// NewKubeClient creates an empty KubeClient
+func NewKubeClient(s *cli.EnvSettings) *kubeClient {
```
sriumcp: Make this `newKubeClient` (don't expose).

autox/config.go
```diff
@@ -7,70 +7,81 @@ import (
...
 
 	"sigs.k8s.io/yaml"
 )
 
-// resourceConfig is the configuration that identifies the resources to watch and in which namespaces.
-type resourceConfig struct {
-	// Namespaces is list of namespaces to watch
-	Namespaces []string `json:"namespaces,omitempty" yaml:"namespaces,omitempty"`
+// Opts are the options used for launching autoX service
+type Opts struct {
```
sriumcp: Since k8sClient is `package visible` variable, we don't need Opts.

autox/watcher.go
```diff
@@ -1,36 +1,57 @@
...
+			break
+		}
+	}
+
+	return err
+}
+
 // Start is entry point to configure services and start them
-func Start(stopCh chan struct{}) error {
+func (opts *Opts) Start(stopCh chan struct{}) error {
```
sriumcp: Since Opts is going away, make `Start` a func.

sriumcp: What if we use the following signature for `Start`.

```go
Start(stopCh chan struct{}, autoxK KubeClientInterface) {
  if autoxK == nil {
    // get the real client
  } else {
    // set it here
    k8sClient = autoxK
  }
}
```

Along  with the above, we can also have a `fake` package.

```go
package fake // goes under autox
// a func to return the fake client
// fake client implements the autox KubeClientInterface
```

cmd/root.go
```diff
@@ -19,8 +20,10 @@ var (
 	logLevel = "info"
 	// Default Helm and Kubernetes settings
 	settings = cli.New()
-	// Kuberdriver used by Helm and Kubernetes clients
+	// KubeDriver used by actions package
 	kd = driver.NewKubeDriver(settings)
+	// KubeClient used by autoX
```
sriumcp:
```go
var autoXKClient autox.KubeClientInterface
```

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+	trackLabel   = "iter8.tools/track"
 )
 
 type chartAction int64
 
 const (
 	releaseAction chartAction = 0
 	deleteAction  chartAction = 1
+
+	applicationTemplateFilePath string = "application.tpl"
```
sriumcp: https://pkg.go.dev/embed

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+	namespace string
+
+	// owner is the chart group secret for this application
+	owner struct {
+		name string
+		uid  string
+	}
+
+	// chart is the Helm chart for this application
+	chart struct {
```
sriumcp: ReleaseSpec?

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+	version := prunedLabels[versionLabel]
+	_, _ = hasher.WriteString(version)
+
+	// hash track label
+	track := prunedLabels[trackLabel]
+	_, _ = hasher.WriteString(track)
+
+	nonce := fmt.Sprintf("%05x", hasher.Sum64())
+	nonce = nonce[:5]
+	return fmt.Sprintf("autox-%s-%s-%s", chartGroupName, chartID, nonce)
```
sriumcp: ReleaseGroupSpecName, ReleaseSpecName, and nonce

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
 }
 
-func newIter8Watcher() *iter8Watcher {
+func newIter8Watcher(k8sClient *kubeClient) *iter8Watcher {
 	w := &iter8Watcher{
 		factories: map[string]dynamicinformer.DynamicSharedInformerFactory{},
 	}
+
+	// aggregate all triggers (namespaces and GVR) from the chartGroupConfig
+	triggers := map[string][]schema.GroupVersionResource{}
```
sriumcp: What is this map key?

sriumcp: `triggers` map has `namespace` as its key and the object GVRs within the namespace that it is watching as its value.

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+
+		namespace := chartGroup.Trigger.Namespace
+		gvr := schema.GroupVersionResource{
+			Group:    chartGroup.Trigger.Group,
+			Version:  chartGroup.Trigger.Version,
+			Resource: chartGroup.Trigger.Resource,
+		}
+
+		// add namespace and GVR to triggers
+		if _, ok := triggers[namespace]; ok {
```
sriumcp: !ok?

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+			triggers[namespace] = []schema.GroupVersionResource{gvr}
+		} else {
+			triggers[namespace] = append(triggers[namespace], gvr)
```
sriumcp: Maybe if else is not needed? Maybe append works with `nil`?
```go
triggers[namespace] = append(triggers[namespace], gvr)
```

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
+		}
+	}
+
 	// for each namespace, resource type configure Informer
-	for _, ns := range iter8ResourceConfig.Namespaces {
+	for ns, gvrs := range triggers {
 		w.factories[ns] = dynamicinformer.NewFilteredDynamicSharedInformerFactory(k8sClient.dynamicClient, 0, ns, nil)
-		for _, gvr := range iter8ResourceConfig.Resources {
+		for _, gvr := range gvrs {
 			informer := w.factories[ns].ForResource(gvr)
```
sriumcp: Let us comment as follows ...

```go
// TBD: optimize the informers by supplying the trigger object name, perhaps through list options
```

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
 
-var deleteObject = func(obj interface{}) {
+func deleteObject(obj interface{}) {
 	log.Logger.Debug("Delete:", obj)
 
 	uObj := obj.(*unstructured.Unstructured)
-	resourceName := uObj.GetName()
-	// example label: iter8.tools/autox-group=hello
-	labels := pruneLabels(uObj.GetLabels())
+	prunedLabels := pruneLabels(uObj.GetLabels())
```
sriumcp: First thing in all the handers would be, use the dynamic client, go to the cluster and fetch that object.

Use obj's typemeta and objectmeta for the above.

sriumcp: Create a lock. Acquire the lock at the start of every handler. And release the lock at the end of each handler.

sriumcp: In the deleteObject function, if you go to the cluster, and find that the trigger object is still present, you return (with a warning).

sriumcp: Delete or create or update of an releaseGroupSpec is an atomic operation. All the releaseSpecs within a releaseGroupSpec will be handled together.

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
-	resourceName := uObj.GetName()
-	// example label: iter8.tools/autox-group=hello
-	labels := pruneLabels(uObj.GetLabels())
 
-	// check if the app name matches the name in the autox label
-	if autoxLabelName, ok := labels[autoxLabel]; ok && resourceName == autoxLabelName {
-		// Release Helm charts
-		installHelmReleases(resourceName)
+	// if there is no autoX label, there is nothing to do
+	labels := uObj.GetLabels()
```
sriumcp: These labels come from the cluster object, not the informer object.

autox/informer.go
```diff
@@ -3,146 +3,361 @@ package autox
...
-	// TODO: select labels important for autoX
+	// we will install Helm releases
+	prunedLabels := pruneLabels(labels)
 
-	return labels
+	_ = installHelmReleases(prunedLabels, uObj.GetNamespace())
 }
 
-var updateObject = func(oldObj, obj interface{}) {
+func updateObject(oldObj, obj interface{}) {
```
sriumcp:
```go
func updateObject(_, obj interface{}) {
    // lock
    // defer unlock
    // get obj from cluster
    // obj exists and has the right labels
    // then, addObject(obj)
    // ! (obj exists and has the right labels)
    // then, deleteObject(obj)
}
```

sriumcp: Refactor so that we don't wastefully go to cluster twice, once in update and once in addObject (or deleteObject).

sriumcp: We want idempotence of add, update, and delete operations.

sriumcp (changes requested):
Idempotence of autox + a few minor naming changes, and visibility changes (for kubeclient).

autox/informer.go
sriumcp: Decide between embed and using an arguCD typed application struct.

sriumcp (approved):

sriumcp (changes requested):
ignore `approve`!

autox/application.tpl
```diff
@@ -0,0 +1,27 @@
...
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+  ownerReferences:
+    - apiVersion: v1
+    kind: Secret
+    name: {{ .owner.name}}
+    uid: {{ .owner.uid}}
+spec:
+  destination:
+    namespace: {{ .trigger.Namespace }}
```
kalantar: Do we expect `metadata.namesapce` and `spec.destination.namespace` to be different? Why are we using different values for them?

autox/config.go
```diff
@@ -7,89 +7,73 @@ import (
...
-	if err != nil {
-		log.Logger.Warnf("unable to read configuration file %s: %s", fp, err.Error())
-		return config // empty configuration
-	}
-
-	log.Logger.Debugf("read configuration\n%s", string(yfile))
+	// Version is the version of the Helm chart
+	// TODO: add version constraint, example: "1.16.X"
+	Version string `json:"version" yaml:"version"`
+}
```
kalantar, line 37: Please create issue for this TODO
Do we also need to consider password/token for access to helm chart? Another issue?

autox/watcher.go
```diff
@@ -1,48 +1,82 @@
...
 
-var iter8ResourceConfig resourceConfig
-var iter8ChartGroupConfig chartGroupConfig
+var k8sClient *kubeClient
+var autoXConfig config
+
+func validateConfig(c config) error {
+	var err error
+
+	for releaseGroupSpecID, releaseGroupSpec := range autoXConfig.Specs {
```
kalantar: We pass in `c config` but loop over global `autoXConfig`. One of these is an error? Probably the use of `autoXConfig` based on the call below.

autox/watcher.go
```diff
@@ -1,48 +1,82 @@
...
+	}
+
+	return err
+}
 
 // Start is entry point to configure services and start them
-func Start(stopCh chan struct{}) error {
+func Start(stopCh chan struct{}, autoxK *kubeClient) error {
+	if autoxK == nil {
+		// get the real client
```
kalantar: maybe "get a **new** client" or "get a **default** client"

autox/k8sclient.go
```diff
@@ -7,16 +7,10 @@ import (
 
 	"helm.sh/helm/v3/pkg/cli"
 
-	// Import to initialize client auth plugins.
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
```
kalantar: Why was this import removed? while it isn't needed more than once, it is essential when using any cloud with authentication

autox/watcher.go
```diff
@@ -1,48 +1,82 @@
...
+
+	err := validateConfig(autoXConfig)
+	if err != nil {
+		return err
+	}
 
-	log.Logger.Debug("chartGroupConfig:", iter8ChartGroupConfig)
+	log.Logger.Debug("config:", autoXConfig)
 
 	w := newIter8Watcher()
```
kalantar: I'm a little surprised that we aren't passing the config to the `newIter8Watcher()` ... and relying on a global variable instead. I suggest refactoring to pass the config and eliminate the global.

autox/informer.go
```diff
@@ -3,134 +3,377 @@ package autox
...
-func deleteHelmRelease(releaseName string, chart chart) {
-	// TODO: check if there is a preexisting Helm release
+// deleteHelmRelease with a given release name
+var deleteHelmRelease = func(releaseName string, group string, namespace string) error {
+	gvr := schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
 
-	// TODO: mutex
-
-	// TODO: delete Helm chart
+	err := k8sClient.dynamic().Resource(gvr).Namespace(namespace).Delete(context.TODO(), releaseName, metav1.DeleteOptions{})
```
kalantar: I'm confused by why delete helm release is implemented a delete of a `Deployment` object. Shouldn't we be deleting the `Application` object ?

kalantar (changes requested):

autox/config.go
Alan-Cha: https://github.com/iter8-tools/iter8/issues/1367
https://github.com/iter8-tools/iter8/issues/1368

autox/watcher.go
Alan-Cha: Yes, thanks for catching this. I will loop over `c` instead. I think `validateConfig(autoXConfig)` has a clearer meaning than `validateConfig()`.

autox/k8sclient.go
Alan-Cha: I am not sure why this was deleted. Maybe it was an unintended mistake. I will restore it.

autox/application.tpl
```diff
@@ -0,0 +1,28 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+  ownerReferences:
+    - apiVersion: v1
+    kind: Secret
+    name: {{ .owner.name}}
```
sriumcp: Please include a space after .name and before }} ; also in the next line.

autox/config.go
```diff
@@ -7,89 +7,73 @@ import (
...
 
-// chartGroup is the configuration of all the Helm charts for a particular experiment group
-// the key is an identifier for the chart
-type chartGroup struct {
-	Charts map[string]chart `json:"charts" yaml:"charts"`
-}
+// releaseSpec points to a particular Helm releaseSpec
+type releaseSpec struct {
+	// RepoURL is the url of the Helm repo
+	RepoURL string `json:"repoURL" yaml:"repoURL"`
```
sriumcp: Please drop the `repoURL string` altogether in the releaseSpec **for now**.

We want to hardcode the repoURL **for now** in all the releaseSpecs (as the Iter8 Hub URL, instead of making this configurable by the user). Mainly for security reasons.

sriumcp (changes requested):
Please see initial comments. More comments to follow.

autox/informer.go
```diff
@@ -3,154 +3,420 @@ package autox
...
+			Version:  releaseGroupSpec.Trigger.Version,
+			Resource: releaseGroupSpec.Trigger.Resource,
+		}
+
+		// add namespace and GVR to triggers
+		triggers[namespace] = append(triggers[namespace], gvr)
+	}
+
+	// mutex is passed to event handler so changes are not concurrent
+	var m sync.Mutex
```
sriumcp: Make mutex package local variable (so it is accessible anywhere inside the package).

autox/watcher.go
```diff
@@ -1,50 +1,82 @@
...
+
+		if releaseGroupSpec.Trigger.Version == "" {
+			err = fmt.Errorf("trigger in spec group \"%s\" does not have a version", releaseGroupSpecID)
+			break
+		}
+
+		if releaseGroupSpec.Trigger.Resource == "" {
+			err = fmt.Errorf("trigger in spec group \"%s\" does not have a resource", releaseGroupSpecID)
+			break
+		}
```
sriumcp, line 46: Create an inverse map between `ns/gvr/name` (i.e., a trigger resource object) and a pair; pair has resourceGroupSpecName, and resourceGroupSpec.
***
If the same trigger is used for different pairs, you will discover it here... and and error out.

autox/informer.go
```diff
@@ -3,154 +3,420 @@ package autox
...
+			log.Logger.Warnf("expected object \"%v\" to have label \"%s\" but none were found", clientU, autoXLabel)
+			return
+		}
+
+		// only install Helm releases if autoX label exists
+		_ = addObjectHelper(clientU, prunedLabels, autoXConfig)
 	}
 }
 
+func deleteObject(ns string, gvr schema.GroupVersionResource, m *sync.Mutex, autoXConfig config) func(obj interface{}) {
```
sriumcp:
```go
func deleteObject(resourceGroupSpecName, autoXConfig) func(obj interface{}) {
 	return func(obj interface{}) {
 	  _ = handle(resourceGroupSpecName, autoXConfig)
 	}
 }

func addObject(resourceGroupSpecName, autoXConfig) func(obj interface{}) {
 	return func(obj interface{}) {
 	  _ = handle(resourceGroupSpecName, autoXConfig)
 	}
 }

func updateObject(resourceGroupSpecName, autoXConfig) func(obj interface{}) {
 	return func(oldObj interface{}, obj interface{}) {
 	  _ = handle(resourceGroupSpecName, autoXConfig)
 	}
 }
```

sriumcp (changes requested):
Please see requested changes.

autox/informer.go
```diff
@@ -3,154 +3,359 @@ package autox
...
+	_, _ = hasher.WriteString(version)
+
+	// hash track label
+	track := prunedLabels[trackLabel]
+	_, _ = hasher.WriteString(track)
+
+	nonce := fmt.Sprintf("%05x", hasher.Sum64())
+	nonce = nonce[:5]
+	return fmt.Sprintf("autox-%s-%s-%s", group, releaseSpecID, nonce)
 }
```
kalantar: I presume that we want the same value for nonce for the same input. This implementation creates a new hasher each time it is called resulting in a different seed. The different seed means the return value might be different. To address this either explictily set the seed to a common value or use a single (global) hasher and call hasher.Reset() before each use.

We have/had the same problem in the the abn lookup code.

autox/informer.go
```diff
@@ -3,154 +3,359 @@ package autox
...
+		chart: struct {
+			name    string
+			values  map[string]interface{}
+			version string
+		}{
+			name:    releaseSpec.Name,
+			values:  releaseSpec.Values,
+			version: releaseSpec.Version,
+		},
+	}
```
kalantar: I saw that we took `repoURL` out of release spec. We are still setting `repoURL: {{ .chart.url }}` in `application.tpl`. What value are we using for `url`? If we are hard coding it, are we doing it here or in the template?

kalantar (changes requested):

autox/informer.go
sriumcp: Actually, this is a good point. We can hardcode this for now in `application.tpl`.

autox/application.tpl
```diff
@@ -0,0 +1,28 @@
...
+  destination:
+    namespace: {{ .namespace }}
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    chart: {{ .chart.name }}
+    helm:
+      values: 
+        {{ .chart.values | toYaml | indent 4 }}
+    repoURL: {{ .chart.url }}
```
sriumcp: Hard code this to Iter8 hub URL here.

autox/informer.go
```diff
@@ -3,154 +3,359 @@ package autox
...
-		for _, gvr := range iter8ResourceConfig.Resources {
+
+		// TBD: optimize the informers by supplying the trigger object name, perhaps through list options
+		for _, gvr := range gvrs {
 			informer := w.factories[ns].ForResource(gvr)
 			informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-				AddFunc:    addObject,
-				UpdateFunc: updateObject,
-				DeleteFunc: deleteObject,
+				AddFunc:    addObject(ns, gvr, autoXConfig),
```
sriumcp: I feel these func signatures still need some fine tuning, based on our slack discussion. Could we please discuss when we meet during one of our dev meetings?

***

The logic is good, but it can be made a bit more robust, simply by watching for each trigger resource. FieldSelectors might help with this.

sriumcp (changes requested):
Please see comments on informer.

autox/informer.go
```diff
@@ -3,154 +3,360 @@ package autox
...
+func deleteObjectHelper(uObj *unstructured.Unstructured, prunedLabels map[string]string, autoXConfig config) error {
+	return deleteHelmReleases(prunedLabels, uObj.GetNamespace(), autoXConfig)
+}
+
 type iter8Watcher struct {
 	factories map[string]dynamicinformer.DynamicSharedInformerFactory
 }
 
-func newIter8Watcher() *iter8Watcher {
+func newIter8Watcher(autoXConfig config) *iter8Watcher {
```
sriumcp, line 332: Please put a comment that says the key is namespace

autox/informer.go
```diff
@@ -3,154 +3,360 @@ package autox
...
+
+		namespace := releaseGroupSpec.Trigger.Namespace
+		gvr := schema.GroupVersionResource{
+			Group:    releaseGroupSpec.Trigger.Group,
+			Version:  releaseGroupSpec.Trigger.Version,
+			Resource: releaseGroupSpec.Trigger.Resource,
+		}
+
+		// add namespace and GVR to triggers
+		triggers[namespace] = append(triggers[namespace], gvr)
```
sriumcp: During start up we check to ensure that different resourceGroupSpecs do not have the same trigger.
***
Within a given namespace, you can still have two triggers with the same gvr. Drop duplicate gvrs within the same namespace.

autox/application.tpl
Alan-Cha: `prune`:

https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/#automatic-pruning

By default (and as a safety mechanism), automated sync will not delete resources when Argo CD detects the resource is no longer defined in Git. To prune the resources, a manual sync can always be performed (with pruning checked). Pruning can also be enabled to happen automatically as part of the automated sync by... setting this option.

@sriumcp @kalantar If this is related to git, then should we get rid of it?

`selfHeal`:

https://argo-cd.readthedocs.io/en/stable/user-guide/auto_sync/#automatic-self-healing

By default, changes that are made to the live cluster will not trigger automated sync. To enable automatic sync when the live cluster's state deviates from the state defined in Git... set this option.

We want the changes to trigger a sync but I'm not sure what how Argo CD's Git usage means in relationship with what we are doing.

autox/informer.go
```diff
@@ -3,154 +3,365 @@ package autox
...
-	// if the pruned labels are different, then update by deleting and reinstalling
-	deleteHelmReleases(resourceName)
-	installHelmReleases(resourceName)
+	// namespace and GVR should already match trigger
+	ns := u.GetNamespace()
+	// Note: GVR is from the release group spec, not available through the obj
+	gvr := getGVR(releaseGroupSpec)
+
+	// always delete Helm releases
+	log.Logger.Debugf("delete Helm releases for release group \"%s\"", releaseGroupSpecName)
```
sriumcp: There are situations when deleting always is not desirable. For e.g., consider if your trigger object is a deployment, and its status has changed. We shouldn't recreate the experiment in this instance.

Perhaps delete if the object no longer exists in the cluster, or if the required autox labels have been removed? If not, simply doChartAction(..., create/update, ...) instead?

sriumcp (changes requested):
Please see a small change requested for delete logic.

autox/informer.go
Alan-Cha: The key is now `releaseGroupSpecName`, right? I think I did add a comment.

autox/informer.go
```diff
@@ -3,154 +3,382 @@ package autox
...
 
-			case deleteAction:
-				deleteHelmRelease(releaseName, chart)
+	log.Logger.Debug("Delete chart:", releaseName)
+	return nil
+}
+
+// doChartAction iterates through a given spec group, and performs action for each spec
+// action can be install or delete
+func doChartAction(prunedLabels map[string]string, chartAction chartAction, namespace string, releaseGroupSpec releaseGroupSpec) error {
```
sriumcp: Instead of prunedLabels, pass in releaseGroupSpecName.

autox/informer.go
```diff
@@ -3,154 +3,382 @@ package autox
...
+	// get group
+	group := prunedLabels[autoXLabel]
+
+	var err error
+	for releaseSpecID, releaseSpec := range releaseGroupSpec.ReleaseSpecs {
+		// get release name
+		releaseName := getReleaseName(group, releaseSpecID, prunedLabels)
```
sriumcp: `getReleaseName` using `releaseGroupSpecName`, and `releaseSpecID`.

sriumcp (changes requested):

autox/watcher_test.go
```diff
@@ -15,7 +15,7 @@ import (
 func TestStart(t *testing.T) {
 	// autoX watcher will call on installHelmRelease
 	installHelmReleaseInvocations := 0
-	installHelmRelease = func(releaseName string, group string, releaseSpec releaseSpec, namespace string) error {
+	installHelmRelease = func(releaseName string, group string, releaseSpec releaseSpec, namespace string, additionapplicationValues map[string]string) error {
```
sriumcp: camelCase `additionalApplicationValues`

sriumcp (changes requested):
Minor comment.

autox/application.tpl
```diff
@@ -0,0 +1,28 @@
...
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    chart: {{ .chart.name }}
+    helm:
+      values: 
+        {{ .chart.values | toYaml | indent 4 }}
+    repoURL: https://iter8-tools.github.io/hub
+    targetRevision: {{ .chart.version }}
+  syncPolicy:
```
sriumcp, line 23: Remove the `syncPolicy` section ... and we will try to see if we still get the ArgoCD behavior we want.

autox/config.go
```diff
@@ -7,89 +7,70 @@ import (
...
-	if err != nil {
-		log.Logger.Warnf("unable to read configuration file %s: %s", fp, err.Error())
-		return config // empty configuration
-	}
+	// Values is the values of the Helm chart
+	Values map[string]interface{} `json:"values" yaml:"values"`
 
-	log.Logger.Debugf("read configuration\n%s", string(yfile))
+	// Version is the version of the Helm chart
+	// TODO: add version constraint, example: "1.16.X"
```
sriumcp: TODO can go.

autox/watcher.go
```diff
@@ -1,50 +1,97 @@
...
+			err = fmt.Errorf("trigger in spec group \"%s\" does not have a name", releaseGroupSpecID)
+			break
+		}
+
+		if releaseGroupSpec.Trigger.Namespace == "" {
+			err = fmt.Errorf("trigger in spec group \"%s\" does not have a namespace", releaseGroupSpecID)
+			break
+		}
+
+		if releaseGroupSpec.Trigger.Group == "" {
```
sriumcp: Remove this check. Group can be "".

base/util.go
```diff
@@ -75,7 +75,7 @@ func getTextTemplateFromURL(providerURL string) (*template.Template, error) {
 		return nil, err
 	}
 
-	tpl, err := createTemplate(string(responseBody))
+	tpl, err := CreateTemplate(string(responseBody))
```
sriumcp, line 85: Lets create an issue ... we want a separate `util` package and we want `base`, `autox` and `abn` -- all of them to depend on it. It is possible that  the util package might absorb the k8s clients in the future.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
-	installHelmReleases(resourceName)
+	// apply Helm releases if (client) object exists and has autoX label
+	// fetch (client) object from cluster
+	// clientU, _ := k8sClient.dynamicClient.Resource(gvr).Namespace(ns).Get(context.TODO(), name, metav1.GetOptions{})
+	if clientU != nil {
+		clientName := clientU.GetName()
+		clientNs := clientU.GetNamespace()
+
+		// sanity check
+		if clientName != name {
```
sriumcp: This is redundant. Please remove.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+			log.Logger.Errorf("autoX expected Kubernetes object to have name \"%s\" but had name \"%s\" instead", name, clientName)
+			return
+		}
+		if clientNs != ns {
```
sriumcp: This is also redundant. Please remove.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
-		for _, gvr := range iter8ResourceConfig.Resources {
-			informer := w.factories[ns].ForResource(gvr)
-			informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-				AddFunc:    addObject,
-				UpdateFunc: updateObject,
-				DeleteFunc: deleteObject,
-			})
-		}
+
+	// aggregate all triggers (namespaces and GVR) from the releaseGroupConfig
```
sriumcp: // we will create a factory for each releaseGroupSpecName. Since there is a 1-1 correspondence between triggers and releaseGroupSpecs, we are effectively creating one factory per trigger.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+
+	// parse object
+	u := obj.(*unstructured.Unstructured)
+
+	// check if name matches trigger
+	name := u.GetName()
+	if name != releaseGroupSpec.Trigger.Name {
+		return
+	}
+
```
sriumcp, line 252: // At this point, we know that we are really handling an event for the trigger object

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/tools/cache"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
-	autoxLabel = "iter8.tools/autox-group"
+	autoXLabel          = "iter8.tools/autox-group"
```
sriumcp:
```go
# label always set to true when present
# Autox controller will only check for existence of this label, not its value
autoXLabel = "iter8.tools/autox"
```

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+		if !hasAutoXLabel(clientLabels) {
+			log.Logger.Debugf("delete Helm releases for release group \"%s\"", releaseGroupSpecName)
 
-	// if the pruned labels are the same, do nothing
-	if reflect.DeepEqual(oldLabels, labels) {
+			_ = doChartAction(deleteAction, releaseGroupSpecName, ns, releaseGroupSpec, nil)
+
+			// if autoX label does not exist, there is no need to apply Helm releases, so return
+			return
+		}
```
sriumcp, line 284: // else do apply action.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+		clientLabels := clientU.GetLabels()
+		clientPrunedLabels := pruneLabels(clientLabels)
+		if !hasAutoXLabel(clientLabels) {
+			log.Logger.Debugf("do not apply Helm releases for release group \"%s\" because Kubernetes object \"%s\" in namespace \"%s\" does not have %s label", releaseGroupSpecName, clientName, clientNs, autoXLabel)
+			return
+		}
+
+		// apply Helm releases
+		log.Logger.Debugf("apply Helm releases for release group \"%s\"", releaseGroupSpecName)
+		_ = doChartAction(applyAction, releaseGroupSpecName, clientNs, releaseGroupSpec, clientPrunedLabels)
```
sriumcp: apply action is now moving up to the else block above.

autox/informer.go
```diff
@@ -3,154 +3,363 @@ package autox
...
+	var buf bytes.Buffer
+	err = tpl.Execute(&buf, values)
+	if err != nil {
+		log.Logger.Error("could not execute application template")
+		return err
+	}
 
-	// TODO: delete Helm chart
+	// serialize application manifest into unstructured object
+	obj, _, err := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme).Decode(buf.Bytes(), nil, nil)
```
sriumcp: This is what we want to use to get an unstructured object out of bytes: https://github.com/kubernetes/client-go/blob/1ac8d459351e21458fd1041f41e43403eadcbdba/dynamic/simple.go#L186

autox/informer_test.go
```diff
@@ -1,51 +1,39 @@
...
-	// Overwrite original handlers
-	addObject = func(obj interface{}) {
-		log.Logger.Debug("Add:", obj)
-		addObjectInvocations++
+	// autoX needs the config
+	autoXConfig := readConfig("../testdata/autox_inputs/config.example.yaml")
+
+	// autoX handler will call on applyHelmRelease and deleteHelmRelease
+	applyHelmReleaseInvocations := 0
+	applyHelmRelease = func(releaseName string, specGroupName string, releaseSpec releaseSpec, namespace string, additionalValues map[string]string) error {
```
sriumcp: Query the cluster. Do not mock `applyHelmRelease` and `delete`. Instead query the cluster and make sure it looks right (e.g., no objects after delete; appropriate application objects after apply).

sriumcp (changes requested):

base/util.go
Alan-Cha: #1378

autox/informer.go
Alan-Cha: @sriumcp I just realized that we are also using `autoXLabel` in a slightly different context here: https://github.com/Alan-Cha/iter8/blob/9ea003f729dd4f144aa77855942079a2a656ad1a/autox/informer.go#L82

I don't think it makes sense for it to be the `autoXLabel` in the secrets to be set to a boolean. I will create a new label called `autoXGroupLabel` for this use case then so that this label doesn't get mixed up with the `autoXLabel` in the trigger objects.

Alan-Cha:
See #1387

PR #1360, opened 2022-10-21
Make istio-prom more flexible

### LINKED ISSUE #1359: `istio-prom` two versions tutorial is not working
There may be a bug with the Istio's Prometheus plug-in where the labels are not set as expected. Our two versions are identified via the `destination_workload` but instead of `httpbin-v1` and `httpbin-v2`, the data on Prometheus is labels as `unknown`.

One work around is to simply to use another set of labels. Another one, is to make the provider template more flexible, instead of being locked to only a few.

Alan-Cha:
Fix #1359


kalantar (dismissed):
looks ok

sriumcp (approved):
lgtm

sriumcp (changes requested):
I have a tangential request in this PR … the golangci-lint.yml workflow  currently lints code not just on PRs, but also on pushes.

```yaml
on:
  push:
    tags:
      - v*
    branches:
      - master
      - main
  pull_request:
```

Could we please change it so that linting happens only during PRs?

```yaml
on:
  pull_request:
```

This is because: 1) the linting is wasteful after merge, and 2) waiting for checks to pass before release is a good thing … but in this case, we are waiting needlessly every time.

sriumcp (approved):
lgtm

PR #1362, opened 2022-11-03
remove post build test gh-action

kalantar:
Remove failing test case. It is in the wrong place for what it is trying to test and the alternative is already being tested.


sriumcp (approved):
lgtm

PR #1372, opened 2022-11-17
fix lookup hash

### LINKED ISSUE #1371: A/B/n hashing not working
**Describe the bug**
It appears that the algorithm used by lookup is not working. In the past, this was tested by deploying a sample application and (after `kubectl port-forward) calling:

    curl localhost:8090/getRecommendation -H 'X-User: foo'

and

    curl localhost:8090/getRecommendation -H 'X-User: foobar'

This method was configured return the version of the backend service that got called, either `v1` or `v2`. It seemed to be working but it would have been easy to miss since each was called just once and the response was so similar.

As part of preparing a blog article on A/B/n testing, the backend service response was changed to be something like: 

    {"Id":19,"Name":"sample","Source":"backend-554c6bbc5b-zpxgb"}

The difference in the last field is more obvious and it was noticed that the the responses were not always different. More importantly, trying the same input repeatedly yields different results. For example:

```shell
% curl localhost:8090/getRecommendation -H 'X-User: foo'
Recommendation: {"Id":19,"Name":"sample","Source":"backend-candidate-7f96fcc99f-55czl"}
% curl localhost:8090/getRecommendation -H 'X-User: foo'
Recommendation: {"Id":19,"Name":"sample","Source":"backend-candidate-7f96fcc99f-55czl"}
% curl localhost:8090/getRecommendation -H 'X-User: foo'
Recommendation: {"Id":19,"Name":"sample","Source":"backend-554c6bbc5b-zpxgb"}
```

**To Reproduce**
Follow instructions [here](https://github.com/kalantar/docs/blob/abn/docs/tutorials/abn/abn.md) to deploy sample service and iter8-abn service. Use curl commands above once it is deployed.

**Expected behavior**
For the same input, the output Source will be the same.

**Additional context**
The problem seems to be the creation of a new hasher each time it is used. This means a different seed is used. Instead, the same hasher should be used and the `Reset()` method should be called.

kalantar:
Fixes https://github.com/iter8-tools/iter8/issues/1371

### DIFF
--- a/abn/core/lookup.go
+++ b/abn/core/lookup.go
@@ -10,6 +10,8 @@ import (
 	abnapp "github.com/iter8-tools/iter8/abn/application"
 )
 
+var versionHasher maphash.Hash
+
 // lookupInternal is detailed implementation of gRPC method Lookup
 func lookupInternal(application string, user string) (*abnapp.Application, *string, error) {
 	// if user is not provided, fail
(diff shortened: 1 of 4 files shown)


sriumcp (changes requested):
The bug fix looks good. Could you also please add a unit test that that exercises this bug? i.e., without the fix, the unit test should fail, but with the fix it succeeds.

sriumcp (approved):
lgtm

PR #1377, opened 2022-11-30
Ability to pin the iter8 install action by digest

### LINKED ISSUE #1376: Make it possible to digest-pin the iter8 install action
It is recommended to pin github actions by their commit hash to ensure reproduciability and enhance the supply chain security: https://michaelheap.com/ensure-github-actions-pinned-sha/. This is also something the OpenSSF Scorecard projects will notify the users about: https://github.com/ossf/scorecard.

It's currently not possible to pin the iter8-tools/iter8 action by its commit hash since the action relies on the tag to determine the version to install: https://github.com/iter8-tools/iter8/blob/master/action.yaml#L8

I think a solution would be to pass the version to install as an action input parameter. This parameter could continue to default to `${{ github.action_ref }}`. But people that would like to pin the action can do so while also specifying a version to install. That's similar to how e.g. the helm-kind-action works: https://github.com/helm/kind-action/blob/main/action.yml#L11

**Describe alternatives you've considered**
An even better solution would be automatically updating the action's default-version on every new release - although this might be challenging/annoying depending on the release process.

**Additional context**
I'd be happy to create a PR for the first proposed solution or help discuss the work required for the second.

chgl:
Tested in a random repository over at: https://github.com/chgl/release-please-test/actions/runs/3587492586/jobs/6037900545
See the changes at https://github.com/chgl/release-please-test/pull/8/files. The second step is expected to fail.

Closes #1376

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -1,32 +1,45 @@
-name: 'Iter8 Install'
-description: 'Install Iter8 CLI'
+name: "Iter8 Install"
+description: "Install Iter8 CLI"
+branding:
+  icon: check-circle
+  color: green
+
 runs:
-  using: 'composite'
+  using: "composite"
   steps:
-    - name: Install latest version of Iter8 CLI
+    - name: Install Iter8 CLI
+      shell: bash
       env:
         VERSION: ${{ github.action_ref }}
-      run: | 
+      run: |
         # For details on context variables runner.os and runner.arch, see:
         #    https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context
-        # The returned values do not match those used by the build artifacts. 
+        # The returned values do not match those used by the build artifacts.
         # See https://github.com/iter8-tools/iter8/blob/master/Makefile#L4 (and for other valid combinations: https://github.com/mitchellh/gox/blob/master/platform.go#L28-L101)
         # Therefore, this code trasforms the context variables as follows:
         #   os: Linux -> linux, Windows -> windows, macOS -> darwin
         #   arch: X86 -> 386 amd X64 -> amd64
         OS=$(echo ${{ runner.os }} | tr '[:upper:]' '[:lower:]' | sed 's/macos/darwin/')
         ARCH=$(echo ${{ runner.arch }} | sed 's/X86/386/' | sed 's/X64/amd64/')
-        
+
         RELEASE="${{ env.VERSION }}"
         if [ -z $RELEASE ] || [ $RELEASE = "stable" ]; then
-          RELEASE=$(curl --silent "https://api.github.com/repos/iter8-tools/iter8/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
-        else
-          RELEASE=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name' | grep $RELEASE | sort -Vr | head -1)
+          REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases/latest" | jq -r '.tag_name')
         fi
-        echo "Installing Iter8 version $RELEASE"
+        if [ -z $REL ]; then
+         REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/tags" | jq --arg RELEASE $RELEASE -r '.[] | select(.commit.sha==$RELEASE) | .name')
+        fi
+        if [ -z $REL ]; then
+          REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name' | grep $RELEASE | head -1)
+        fi
+        if [ -z $REL ]; then
+          echo "Invalid release specified: $RELEASE"
+          exit 1
+        fi
+        echo "Installing Iter8 version $REL"
 
         ASSET="iter8-$OS-$ARCH.tar.gz"
-        ASSET_URL="https://github.com/iter8-tools/iter8/releases/download/$RELEASE/$ASSET"
+        ASSET_URL="https://github.com/iter8-tools/iter8/releases/download/$REL/$ASSET"
         echo "Downloading $ASSET_URL"
         wget -q $ASSET_URL && rc=$? || rc=$?
         if [ $rc -eq 0 ]; then
@@ -37,7 +50,3 @@ runs:
           echo "ERROR: unable to download $ASSET from $ASSET_URL"
           exit 1
         fi
-      shell: bash
-branding:
-  icon: check-circle
-  color: green


action.yaml
```diff
@@ -1,22 +1,27 @@
...
+    required: true
+    default: "${{ github.action_ref }}"
 runs:
   using: 'composite'
   steps:
     - name: Install latest version of Iter8 CLI
       env:
-        VERSION: ${{ github.action_ref }}
-      run: | 
+        VERSION: ${{ inputs.version }}
```
sriumcp: Is `env.VERSION` still required? Should we replace `env.VERSION` with `inputs.version`?

Also, I just tried 
```shell
curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name'
```

The `action.yaml` file seems to run the above when a SHA is supplied to `inputs.version`.

I get back the following (version tags; no SHA) ... 
```shell
v0.12.2
v0.12.0
v0.12.2-pre
v0.12.1-pre
v0.12.0-pre
v0.11.16
v0.11.15
v0.11.14
v0.11.13
v0.11.12
v0.11.11
v0.11.10
v0.11.9
v0.11.8
v0.11.7
v0.11.6
v0.11.5
v0.11.4
v0.11.3
v0.11.2
v0.11.1
v0.11.0
v0.10.25
v0.10.24
v0.10.23
v0.10.22
v0.10.21
v0.10.20
v0.10.19
v0.10.18
```

Should `action.yaml` change along the below lines to make the SHA pinning work (very rough attempt below)?

```shell
# handle the case when no inputs.version is specified or when inputs.version is set to (latest) stable
if [ -z ${{ inputs.version}} ] || [ ${{ inputs.version}} == "stable" ]
then
  RELEASE=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases/latest" | jq .tag_name)
# TBD: figure out the regex for the elif below
elif # ${{ inputs.version}} matches a tag (e.g., `v0.12` or `v0.12.2`)
then
  RELEASE=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name' | grep ${{ inputs.version }} | head -1)
else # ${{ inputs.version}} has to be a SHA at this point
  RELEASE=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/tags" | jq '.[] | select(.commit.sha=="${{inputs.version}}") | .name')
fi
```

sriumcp: Also, we have gone back-and-forth on the following idea. But, SHA pinning of Iter8 action seems like a good reason to have a separate repo for the iter8 action altogether. Then we can ... 

```yaml
-uses: iter8-tools/iter8-action@<sha commit>
```

And within the `action.yaml`, we can fix the Iter8 release that is getting used to install Iter8!

FYI @kalantar.

kalantar:
The logic to find a release will require modification as @sriumcp suggests. However, the code to download the iter8 binary depends on the release as well. Can we download the asset by SHA?

kalantar:
The logic to select the release might be simplied (remove the then) as:
```
if [ -z $RELEASE ] || [ $RELEASE = "stable" ]; then
  REL=$(curl --silent "https://api.github.com/repos/iter8-tools/iter8/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
fi
if [ -z $REL ]; then
  REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name' | grep -w $RELEASE | sort -Vr | head -1)
fi
if [ -z $REL ]; then
 REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/tags" | jq -r --arg RELEASE $RELEASE '.[] | select(.commit.sha==$RELEASE) | .name')
fi
if [ -z $REL ]; then
  echo "Invalid release specified: $RELEASE"
  exit 1
fi
echo "Installing Iter8 version $REL"
```

action.yaml
chgl: > Is env.VERSION still required? Should we replace env.VERSION with inputs.version?

Not necessarily, I can definitely replace env.VERSION - the current way just keeps the number of changes minimal. Should I?

> Should action.yaml change along the below lines to make the SHA pinning work (very rough attempt below)?

I think that's a really good approach - instead of finding the release from its tag, find the tag from the commit sha. It's generally possible that one might specify a commit sha that doesn't correspond to a git tag (and therefore a release), however most tools to auto-update a digest-pinned action will follow git tags anyways so it shouldn't be a problem.

sriumcp: Combining all of the above discussion with [this suggestion](https://github.com/iter8-tools/iter8/pull/1377#issuecomment-1333988464) ... could we please modify the `action.yaml` in this PR so that it looks as follows. I realize the changes below are more than minimal, but SHA pinning is clearly well worth achieving, and this seems to get us there.

```yaml
name: 'Iter8 Install'
description: 'Install Iter8 CLI'
runs:
  using: 'composite'
  steps:
    - name: Install Iter8 CLI
      env:
        VERSION: ${{ github.action_ref }}
      run: | 
        # For details on context variables runner.os and runner.arch, see:
        #    https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context
        # The returned values do not match those used by the build artifacts. 
        # See https://github.com/iter8-tools/iter8/blob/master/Makefile#L4 (and for other valid combinations: https://github.com/mitchellh/gox/blob/master/platform.go#L28-L101)
        # Therefore, this code trasforms the context variables as follows:
        #   os: Linux -> linux, Windows -> windows, macOS -> darwin
        #   arch: X86 -> 386 amd X64 -> amd64
        OS=$(echo ${{ runner.os }} | tr '[:upper:]' '[:lower:]' | sed 's/macos/darwin/')
        ARCH=$(echo ${{ runner.arch }} | sed 's/X86/386/' | sed 's/X64/amd64/')
        
        RELEASE="${{ env.VERSION }}"
        if [ -z $RELEASE ] || [ $RELEASE = "stable" ]; then
          REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases/latest" | jq -r '.tag_name')
        fi
        if [ -z $REL ]; then
         REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/tags" | jq --arg RELEASE $RELEASE -r '.[] | select(.commit.sha==$RELEASE) | .name')
        fi
        if [ -z $REL ]; then
          REL=$(curl -s "https://api.github.com/repos/iter8-tools/iter8/releases" | jq -r '.[] | .tag_name' | grep $RELEASE | head -1)
        fi
        if [ -z $REL ]; then
          echo "Invalid release specified: $RELEASE"
          exit 1
        fi
        echo "Installing Iter8 version $REL"

        ASSET="iter8-$OS-$ARCH.tar.gz"
        ASSET_URL="https://github.com/iter8-tools/iter8/releases/download/$REL/$ASSET"
        echo "Downloading $ASSET_URL"
        wget -q $ASSET_URL && rc=$? || rc=$?
        if [ $rc -eq 0 ]; then
          # echo "SUCCESS: downloaded $ASSET from $ASSET_URL"
          tar -xvf $ASSET
          mv $OS-$ARCH/iter8 /usr/local/bin
        else
          echo "ERROR: unable to download $ASSET from $ASSET_URL"
          exit 1
        fi
      shell: bash
branding:
  icon: check-circle
  color: green
```

FYI... @kalantar

Some finer points:

1. I removed the `sort` in the above commands -- the GitHub APIs already list results in the reverse chronological order. The sort was breaking this (and making earlier pre-release versions show at up at the top of the list).
2. I further simplified the first `if` where we are fetching the latest release.

sriumcp (changes requested):

chgl:
Sounds good to me! I'll get to it in a couple hours. Really like the solution!

sriumcp (approved):
lgtm

kalantar (approved):
lgtm

PR #1389, opened 2023-01-20
make dockerfile single stage; add platforms

kalantar:
We no longer need a multi-stage docker build. This simplifies it.
The image we build does not easily run on Mac M1 architecture. Explicitly identifying a build platform helps.


sriumcp (approved):
lgtm

PR #1391, opened 2023-01-24
fix update informers

kalantar:
Fix problem with update handler that manifested when objects deleted (ie, when finalizer should be removed)

### DIFF
--- a/abn/watcher/informer.go
+++ b/abn/watcher/informer.go
@@ -82,6 +82,7 @@ func NewIter8Watcher(configFile string) *Iter8Watcher {
 
 		// create informer for each resource in namespace
 		for gvr, validNames := range byResource {
+			validNames := validNames
 			gvr := gvr
 			informer := w.factories[ns].ForResource(gvr)
 			_, err := informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
--- a/abn/watcher/informer.go
+++ b/abn/watcher/informer.go
@@ -18,6 +18,12 @@ type Iter8Watcher struct {
 	factories map[string]dynamicinformer.DynamicSharedInformerFactory
 }
 
+const (
+	addAction    = "ADD"
+	updateAction = "UPDATE"
+	deleteAction = "DELETE"
+)
+
 // NewIter8Watcher returns a watcher for iter8 related objects
 func NewIter8Watcher(configFile string) *Iter8Watcher {
 	c := readServiceConfig(configFile)
@@ -26,11 +32,29 @@ func NewIter8Watcher(configFile string) *Iter8Watcher {
 		factories: map[string]dynamicinformer.DynamicSharedInformerFactory{},
 	}
 
-	handlerFunc := func(validNames []string, gvr schema.GroupVersionResource) func(obj interface{}) {
+	addHandlerFunc := func(validNames []string, gvr schema.GroupVersionResource) func(obj interface{}) {
 		return func(obj interface{}) {
 			o := obj.(*unstructured.Unstructured)
 			if containsString(validNames, o.GetName()) {
-				handle(o, c, w.factories, gvr)
+				handle(o, c, w.factories, gvr, addAction)
+			}
+		}
+	}
+
+	updateHandlerFunc := func(validNames []string, gvr schema.GroupVersionResource) func(oldObj, obj interface{}) {
+		return func(oldObj, obj interface{}) {
+			o := obj.(*unstructured.Unstructured)
+			if containsString(validNames, o.GetName()) {
+				handle(o, c, w.factories, gvr, updateAction)
+			}
+		}
+	}
+
+	deleteHandlerFunc := func(validNames []string, gvr schema.GroupVersionResource) func(obj interface{}) {
+		return func(obj interface{}) {
+			o := obj.(*unstructured.Unstructured)
+			if containsString(validNames, o.GetName()) {
+				handle(o, c, w.factories, gvr, deleteAction)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


sriumcp (commented):
The `handle` doesn't actually do anything different based on the action type. So, why introduce the action type at all?

Also, I don't see any lock or unlock (which is something `autox` implements). Essentially, two different handle methods for the same application should not be invoked simultaneously. One of them must release the lock, so that the other can acquire the lock and proceed. 

Is the real problem `lock/unlock` and is the `action` type hiding the real issue?

kalantar:
> The handle doesn't actually do anything different based on the action type. So, why introduce the action type at all?

True. It is used in trace level debug statements only. I can remove it. I had it in the past and had removed it. I found it useful again. This time I left it. I can remove it if you feel this is best.

kalantar:
> Also, I don't see any lock or unlock (which is something autox implements). Essentially, two different handle methods for the same application should not be invoked simultaneously.

The autox model bothers me — it is essentially serializing all responses on all objects. I think we want to do so only on an application by application basis. The abn service does this when updating its internal application representation from the objects read from the cluster.

kalantar:
> Is the real problem lock/unlock and is the action type hiding the real issue?

The real problem was how the handler was defined. It was defined incorrectly essentially making the gvr fixed no matter the object.

sriumcp:
> > The handle doesn't actually do anything different based on the action type. So, why introduce the action type at all?
> 
> True. It is used in trace level debug statements only. I can remove it. I had it in the past and had removed it. I found it useful again. This time I left it. I can remove it if you feel this is best.

It is ok to have the `add/update/deleteHandlerFunc(s)`. As you said, you can do better logging this way. But it would be preferable not to pass the `action type` into the `handle` function. We want the `handle` invocation to be identical across the different types of events (similar to how a `reconcile` method in a controller is identical in every type of event).

Passing in the `event type` into `handle` could, in principle, enable `handle` implementations in the future to make use of that argument, which is what we want to avoid (in the spirit of declarative).

sriumcp:
> The autox model bothers me — it is essentially serializing all responses on all objects. I think we want to do so only on an application by application basis. The abn service does this when updating its internal application representation from the objects read from the cluster.

Yes, this design choice in autox assumes that changes to Kubernetes resource objects are infrequent. If this turns out to be not true (e.g., a user complains about performance), we can certainly introduce application level locks in autox also. Meanwhile, the autox code base is simpler (which also means, it is easier to catch bugs during reviews / maintain).

sriumcp:
> > Is the real problem lock/unlock and is the action type hiding the real issue?
> 
> The real problem was how the handler was defined. It was defined incorrectly essentially making the gvr fixed no matter the object.

How does the PR fix this issue?

kalantar:
> Passing in the event type into handle could, in principle, enable handle implementations in the future to make use of that argument, which is what we want to avoid (in the spirit of declarative).

I've removed the parameter. I moved all the logging so that it isn't needed.

kalantar:
>>> Is the real problem lock/unlock and is the action type hiding the real issue?

>> The real problem was how the handler was defined. It was defined incorrectly essentially making the gvr fixed no matter the object.

> How does the PR fix this issue?

The common method handle() deals with events of all types.

The original definition of the add/delete handler was (provided by handerFunc()) a function that calls handle(). The values of the parameters to handle (except obj) are fixed at informer defintion time.

The original definition of the update handler was incorrect. It was a a function that called handlerfFunc() which then called called handle(). The values of the parameters to handle are then set incorrectly.

The update handler function definition was fixed. It should be clearer that it is correct since there is now a handlerFunc for each event type.  In the past this wasn't an issue. The gvr was not passed by derived at runtime from the object. The other parameters are common to all events so did not matter.

sriumcp:
> > > > Is the real problem lock/unlock and is the action type hiding the real issue?
> 
> > > The real problem was how the handler was defined. It was defined incorrectly essentially making the gvr fixed no matter the object.
> 
> > How does the PR fix this issue?
> 
> The common method handle() deals with events of all types.
> 
> The original definition of the add/delete handler was (provided by handerFunc()) a function that calls handle(). The values of the parameters to handle (except obj) are fixed at informer defintion time.
> 
> The original definition of the update handler was incorrect. It was a a function that called handlerfFunc() which then called called handle(). The values of the parameters to handle are then set incorrectly.
> 
> The update handler function definition was fixed. It should be clearer that it is correct since there is now a handlerFunc for each event type. In the past this wasn't an issue. The gvr was not passed by derived at runtime from the object. The other parameters are common to all events so did not matter.

So .... is this a golang loop variable issue? https://github.com/golang/go/discussions/56010

If so, should the `gvr` be redefined inside the for loop?

```go
gvr := gvr
```

If it is a loop variable issue, I suspect the problem is still not solved.

abn/watcher/informer.go
```diff
@@ -55,13 +82,12 @@ func NewIter8Watcher(configFile string) *Iter8Watcher {
 
 		// create informer for each resource in namespace
 		for gvr, validNames := range byResource {
+			gvr := gvr
```
sriumcp, line 86: Please also add `validNames := validNames`. Note that both the loop variables are used in a closure.

sriumcp (changes requested):

sriumcp (approved):
lgtm

sriumcp (changes requested):
Could you please `testdata/experiment.yaml` by making the following change?

```yaml
- task: assess
    with:
      SLOs:
        upper:
        - metric: "http/error-rate"
          limit: 0
        - metric: "http/latency-mean"
          limit: 5000
        - metric: "http/latency-p50"
          limit: 5000
        - metric: "http/latency-p50.0"
          limit: 5000
        - metric: "http/latency-p95.0"
          limit: 5000
        - metric: "http/latency-p99"
          limit: 5000
```

sriumcp (approved):
lgtm

PR #1397, opened 2023-02-10
update workflow action versions

kalantar:

### DIFF
--- a/.github/workflows/assets.yaml
+++ b/.github/workflows/assets.yaml
@@ -13,7 +13,7 @@ jobs:
       uses: actions/setup-go@v3
       with:
         go-version: 1.19
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Build binaries
       run: |
         VERSION=${GITHUB_REF#refs/*/}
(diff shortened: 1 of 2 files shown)


kalantar:
We continue to have a problem with the staticcheck job. The action dominikh/staticcheck-action](https://github.com/dominikh/staticcheck-action)@v1.2.0 uses the deprecated `actions/cache@v2` internally. While the source for the action has been updated, there has been no new release.

We can either continue to use for now or comment out the check (we are using the staticcheck plugin in our lint checking as well).

sriumcp (dismissed):
LGTM

sriumcp:
> We continue to have a problem with the staticcheck job. The action dominikh/staticcheck-action](https://github.com/dominikh/staticcheck-action)@v1.2.0 uses the deprecated `actions/cache@v2` internally. While the source for the action has been updated, there has been no new release.
> 
> We can either continue to use for now or comment out the check (we are using the staticcheck plugin in our lint checking as well).

Great catch. Lets get rid of it since `golanglint-ci` covers static check also.

sriumcp (approved):
LGTM

Alan-Cha (approved):

PR #1409, opened 2023-02-27
Multiple endpoints

Alan-Cha:


base/collect_http_test.go
```diff
@@ -1,32 +1,43 @@
...
 	_ = os.Chdir(t.TempDir())
 	httpmock.Activate()
 	t.Cleanup(httpmock.DeactivateAndReset)
 
-	// Exact URL match
-	httpmock.RegisterResponder("POST", "https://something.com",
-		httpmock.NewStringResponder(200, `[{"id": 1, "name": "My Great Thing"}]`))
+	// // Exact URL match
+	// httpmock.RegisterResponder("POST", "https://something.com",
+	// 	httpmock.NewStringResponder(200, `[{"id": 1, "name": "My Great Thing"}]`))
```
Alan-Cha: These fortio tests are not actually using the mocked endpoints.

base/collect_grpc.go
```diff
@@ -64,25 +72,72 @@ func (t *collectGRPCTask) validateInputs() error {
...
-			e = errors.New("failed to get results since ghz run was aborted")
-			log.Logger.Error(e)
+			// merge endpoint options with baseline options
+			if err := mergo.Merge(&endpoint, t.With.Config); err != nil {
+				log.Logger.Error(fmt.Sprintf("could not merge Fortio options for endpoint \"%s\"", endpointID))
+				return nil, err
+			}
+			eOpts := runner.WithConfig(&endpoint) // endpoint options
+
+			igr, err := runner.Run(call, host, eOpts)
```
kalantar, line 105: Are we accessing the endpoints sequentially? Is this what we want?  Or do we want this to be concurrent executions? I think this would be my expectation.

base/collect_grpc.go
```diff
@@ -64,25 +72,72 @@ func (t *collectGRPCTask) validateInputs() error {
...
 // resultForVersion collects gRPC test result for a given version
-func (t *collectGRPCTask) resultForVersion() (*runner.Report, error) {
+func (t *collectGRPCTask) resultForVersion() (map[string]*runner.Report, error) {
 	// the main idea is to run ghz with proper options
 
-	opts := runner.WithConfig(&t.With.Config)
+	var err error
+	results := map[string]*runner.Report{}
+
+	if len(t.With.Endpoints) > 0 {
```
kalantar, line 81: would it make sense to "create an endpoint" if there is none. Then processing is the same; no need for conditional run code.

base/collect_http.go
```diff
@@ -20,8 +21,8 @@ type errorRange struct {
 	Upper *int `json:"upper,omitempty" yaml:"upper,omitempty"`
 }
 
-// collectHTTPInputs contain the inputs to the metrics collection task to be executed.
-type collectHTTPInputs struct {
+// collectHTTPInputsHelper contains the inputs for one endpoint
+type collectHTTPInputsHelper struct {
```
kalantar: Perhaps rename to make clear. `Endpoint` ?

base/collect_http.go
```diff
@@ -210,24 +219,70 @@ func (t *collectHTTPTask) getFortioOptions() (*fhttp.HTTPRunnerOptions, error) {
...
+			endpoint := endpoint // prevent implicit memory aliasing
+			log.Logger.Trace(fmt.Sprintf("endpoint: %s", endpointID))
+
+			// merge endpoint config with baseline config
+			if err := mergo.Merge(&endpoint, t.With.collectHTTPInputsHelper); err != nil {
+				log.Logger.Error(fmt.Sprintf("could not merge Fortio options for endpoint \"%s\"", endpointID))
+				return nil, err
+			}
+
+			efo, err := getFortioOptions(endpoint)
```
kalantar, line 241: same question about concurrency as for grpc

base/collect_http.go
```diff
@@ -210,24 +219,70 @@ func (t *collectHTTPTask) getFortioOptions() (*fhttp.HTTPRunnerOptions, error) {
...
-	log.Logger.Trace("got fortio options")
-	log.Logger.Trace("URL: ", fo.URL)
-	ifr, err := fhttp.RunHTTPTest(fo)
-	if err != nil {
-		log.Logger.WithStackTrace(err.Error()).Error("fortio failed")
-		if ifr == nil {
-			log.Logger.Error("failed to get results since fortio run was aborted")
+	var err error
+	results := map[string]*fhttp.HTTPRunnerResults{}
+	if len(t.With.Endpoints) > 0 {
```
kalantar, line 229: same comment about creating endpoint as for grpc

base/collect_grpc.go
sriumcp: This is an optimization. Definitely, a useful optimization... but perhaps, one that can handled in a future PR? For now, sequential seems like a good (MVP) start.

base/collect_grpc.go
sriumcp: This is a good implementation choice. Perhaps, given that we already have this code now, we can merge (with a note/issue re: this for future consideration)?

Also, it is unclear what the simplification would be ... this will introduce a condition that says, `if no endpoint, then create one`!

Alan-Cha: @kalantar Are you suggesting that during preprocessing, we will check if there is `Endpoints`, if not and if there is base configuration, then we will add that as an element to `Endpoints`. Then during the `run()` portion, we will not need this check and we can just iterate over `Endpoints`?

sriumcp (approved):
lgtm

kalantar (approved):

Issue #1414, opened 2023-03-09
Explore A/B/n with rewards


kalantar:
Introduce reward  and winner to ABn experiments.

**Use Case**
Assume that metrics are generated by application and written to metrics service such as Prometheus.

*Scenario*
1. Deploy two versions of an application/ML model
2. Run Iter8 experiment specifying reward metrics
3. After each loop, report winner based on reward metrics

*Comments*
- Can also specify SLOs to validate
- A reward metric can also be used to specify an SLO
- Initially report winner for each reward metric separately; ie, not "combined" winner

**Example**
Deploy 2 ML models, for example as in this [canary experiment tutorial](https://github.com/sriumcp/kserve/blob/iter8/docs/samples/canary-testing/README.md).

Run an experiment:
```shell
iter8 k launch \
--set "tasks={ready,custommetrics,assess}" \
--set ready.isvc=${ISVC} \
--set ready.timeout=600s \
--set custommetrics.templates.kserve-prometheus="http://kserve-prometheus.tpl" \
--set custommetrics.values.labels.service_name=${ISVC}-predictor-default \
--set "custommetrics.versionValues[0].labels.revision_name=${CURRENT_REVISION}" \
--set "custommetrics.versionValues[1].labels.revision_name=${CANARY_REVISION}" \
--set "custommetrics.values.latencyPercentiles={50,75,90,95}" \
--set "assess.rewards.greater={kserve-prometheus/conversion-rate,kserve-prometheus/profit}" \
--set assess.SLOs.lower.kserve-prometheus/conversion-rate=.35 \
--set assess.SLOs.upper.kserve-prometheus/latency-mean=50 \
--set assess.SLOs.upper.kserve-prometheus/latency-p90=75 \
--set assess.SLOs.upper.kserve-prometheus/error-count=0 \
--set runner=cronjob \
--set cronjobSchedule="*/1 * * * *"
```

`iter8 k report` should be modified to include information about reward metrics/winners. Perhaps like this:

```
Experiment summary:
*******************

Experiment completed: true
No task failures: true
Total number of tasks: 1
Number of completed tasks: 1
Number of completed loops: 3

Reward evaluation:
*****************

Reward          | winner |
------          | ------ |
conversion-rate | backend-candidate-1 (v2)
profit          | insufficient data

Whether or not service level objectives (SLOs) are satisfied:
*************************************************************

  SLO Conditions                             | backend (v1) | backend-candidate-1 (v2)
[...]

PR #1418, opened 2023-03-21
assess rewards and report

kalantar:
Part of https://github.com/iter8-tools/iter8/issues/1414


sriumcp (commented):
Looking good!

sriumcp (changes requested):
Please include changes to `htmlreporter` also. We want rewards in both text and html.

sriumcp (approved):
lgtm

sriumcp (changes requested):
Could we please run `iter8` as nonRoot (and set user id) in the experiment chart? This will be needed in both job and cronjob templates.

kalantar:
> Could we please run `iter8` as nonRoot (and set user id) in the experiment chart? This will be needed in both job and cronjob templates.

The only writable files for non-root users are in `/tmp`. Moved all downloads to `/tmp' in templates.

sriumcp (approved):
lgtm

PR #1422, opened 2023-03-27
Increase test reliability

kalantar:
This fixes issues with test reliability; test that use httpmock and fortio are unreliable.


sriumcp (approved):
lgtm

PR #1424, opened 2023-03-28
SSA controller

sriumcp:
Handles 4 out of the 5 behaviors in #1427


Alan-Cha (dismissed):

controllers/k8sclient/simple.go
```diff
@@ -0,0 +1,74 @@
...
+	// typed Kubernetes client
+	*kubernetes.Clientset
+	// dynamic Kubernetes client
+	*dynamic.DynamicClient
+}
+
+// Patch performs a server-side apply of GVR
+func (cl *Client) Patch(gvr schema.GroupVersionResource, objNamespace string, objName string, jsonBytes []byte) (*unstructured.Unstructured, error) {
+	return cl.DynamicClient.Resource(gvr).Namespace(objNamespace).Patch(context.TODO(), objName, types.ApplyPatchType, jsonBytes, metav1.PatchOptions{
+		FieldManager: "iter8-controller",
```
kalantar: constant for FieldManager?

cmd/controllers.go
```diff
@@ -0,0 +1,59 @@
...
+				log.Logger.Error("controllers did not start ... ")
+				return err
+			}
+			log.Logger.Debug("started controllers ... ")
+
+			// sigCh := make(chan os.Signal, 1)
+			// signal.Notify(sigCh, syscall.SIGTERM, os.Interrupt)
+			// <-sigCh
+
+			// log.Logger.Warn("SIGTERM ... ")
```
kalantar: Why comment this out? If here, then perhaps also in abn and autox. This is standard to gracefully shutdown when interrupted. Perhaps only relevant when debugging; running locally.

controllers/allcontrollers.go
```diff
@@ -0,0 +1,224 @@
...
+	iter8WatchLabel   = "iter8.tools/watch"
+	iter8WatchValue   = "true"
+
+	// for routemap resource
+	iter8ManagedByLabel    = "app.kubernetes.io/managed-by"
+	iter8ManagedByValue    = "iter8"
+	iter8KindLabel         = "iter8.tools/kind"
+	iter8KindRoutemapValue = "routemap"
+	iter8VersionLabel      = "iter8.tools/version"
+	iter8VersionValue      = "v0.14"
```
kalantar: This is the same as `base.MajorMinor`. We should use that instead or refer to that. In any case, should be **v0.13**.

controllers/allcontrollers.go
```diff
@@ -0,0 +1,224 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/informers/internalinterfaces"
+	"k8s.io/client-go/tools/cache"
+)
+
+const (
+	// for application resources
+	iter8FinalizerStr = "iter8.tools/finalizer"
```
kalantar: A nit, but I would define this in `finalizer.go` where it is used.

controllers/routemaps.go
```diff
@@ -0,0 +1,90 @@
...
+	u := obj.(*unstructured.Unstructured)
+
+	// attempt to return the routemap
+	// ToDo: speed up this quadruple-nested for loop
+	for _, rmByName := range s.nsRoutemap {
+		for _, rm := range rmByName {
+			for _, v := range rm.Variants {
+				for _, r := range v.Resources {
+					if r.GVRShort == gvrShort && r.Name == u.GetName() {
+						if r.Namespace == nil || *r.Namespace == u.GetNamespace() {
```
kalantar, line 37: For this check, r.Namespace == nil means that this resource in the variant was listed w/o a namespace. 
I suppose that in a namespace scoped deployment, a nil ns makes sense. But what does it mean in a cluster scoped deployment? Would it mean the (kubernetes context) default or every ns?  Doesn't it make more sense to require a namespace for all resources?  in a cluster scoped deployment and to set it when it is a namespaced deployment so that at this point there is always a value?

controllers/routemap.go
```diff
@@ -0,0 +1,420 @@
...
+	// routemap CM must be immutable
+	if confMap.Immutable == nil || !(*confMap.Immutable) {
+		err := errors.New("routemap CM is not immutable")
+		log.Logger.Error(err)
+		return err
+	}
+	return nil
+}
+
+// validateRoutemap validates a given sbject
```
kalantar: update comment: remove the word `sbject`

controllers/routemap.go
```diff
@@ -0,0 +1,420 @@
...
+func validateRoutemap(s *routemap, config *Config) (*routemap, error) {
+	// routemap must have at least one variant
+	if len(s.Variants) == 0 {
+		e := errors.New("routemap must at least one variant")
+		log.Logger.Error(e)
+		return nil, e
+	}
+
+	// if !clusterScoped, variant resource namespace should be nil or equal routemap namespace
```
kalantar, line 386: And if it is cluster scoped shouldn't we be checking that a namespace is always set?

controllers/podname.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// getLeaderName returns the name of the leader pod in this statefulset
+// leader pod is the pod with index 0
+func getLeaderName() (string, bool) {
+	statefulSetName, ok := getStatefulSetName()
+	if !ok {
+		return "", false
+	}
+	return statefulSetName + statefulSetSep + "0", true
```
kalantar: why not just use `leaderSuffix` instead of `statfulSetSep + "0"`?

controllers/podname.go
```diff
@@ -0,0 +1,72 @@
...
+	if len(slice) < 2 {
+		return "", false
+	}
+	slice = slice[:len(slice)-1]
+	return strings.Join(slice, statefulSetSep), true
+}
+
+// getLeaderName returns the name of the leader pod in this statefulset
+// leader pod is the pod with index 0
+func getLeaderName() (string, bool) {
```
kalantar: Except in the test cases, I don't see a use of this method. Delete it?  Consequently, `getStatefulSetName()` is equally unused.

controllers/config.go
```diff
@@ -0,0 +1,76 @@
...
+	"path/filepath"
+
+	"github.com/iter8-tools/iter8/base/log"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+const (
+	// ConfigEnv is the name of environment variable with config file path
+	ConfigEnv = "CONFIG_FILE"
```
kalantar: Does this need to be externalized?

controllers/podname.go
```diff
@@ -0,0 +1,72 @@
...
+	"strings"
+
+	"github.com/iter8-tools/iter8/base/log"
+)
+
+const (
+	// PodNameEnvVariable is the name of the environment variable with pod name
+	PodNameEnvVariable = "POD_NAME"
+	// PodNamespaceEnvVariable is the name of the environment variable with pod namespace
+	PodNamespaceEnvVariable = "POD_NAMESPACE"
```
kalantar: Do these need to be externalized?

controllers/routemap.go
```diff
@@ -0,0 +1,420 @@
...
+/* types: end */
+
+const (
+	routemapStrSpec      = "strSpec"
+	weightAnnotation     = "iter8.tools/weight"
+	defaultVariantWeight = uint32(1)
+)
+
+// Weights provide the relative weights for traffic routing between variants
+func (s *routemap) Weights() []uint32 {
```
kalantar, line 73: What is the purpose of this function?  I am unclear on the use case to externalize it.

kalantar (changes requested):
Mostly nits. But a few small issues.

kalantar:
I have a general question about the content of testdata/controllers. We have a number of folders with scripts for various kserve use cases. None of this serves the purpose of test data. I am wondering if this is the right place for this. If so, shouldn't the modelmesh versions also be included (there is a PR on the branch for this).  I think the right place is probably in the docs project under samples.

testdata/controllers/config.yaml
```diff
@@ -0,0 +1,23 @@
+defaultResync: 30s
+# by default, Iter8 controller is namespace scoped
+# clusterScoped: true
+resourceTypes:
```
kalantar, line 4: This needs to include the resource type `cm`. See PR on the branch.

kalantar (changes requested):

testdata/controllers/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    Group: serving.kserve.io
+    Version: v1beta1
+    Resource: inferenceservices
+    conditions:
+    - name: Ready
+      status: "True"
+  vs:
+    Group: networking.istio.io
+    Version: v1beta1
+    Resource: virtualservices
```
kalantar: @Alan-Cha points out that we had initial suggested that `group`, `version` and `resource` should be lower case. I observe that the controller code is independent of case. However, the helm chart seems to be dependent. Clearly, it would be good to make a helm chart that handles both. In the meantime, can we make these lower case?

kalantar (changes requested):

controllers/k8sclient/simple.go
sriumcp: Fixed.

cmd/controllers.go
sriumcp: Fixed in such a way that it enables unit testing of the command itself. Please see comments in code.

controllers/allcontrollers.go
sriumcp: Fixed.

controllers/allcontrollers.go
sriumcp: Fixed.

controllers/routemaps.go
sriumcp: 1. This function is about mapping a resource back to its routemap, and is now well documented.
2. Routemap.variants[].Resource[].Namespace is also now well documented in the type definition.
3. The following func validates the routemap as documented in 2 above. All three are separate but related things.
4. In both cluster-scoped, and namespace scoped modes, the namespace can be omitted; if so, the resource namespace is interpreted to be the routemap namespace as documented now in 2.

```go
// validateRoutemap validates a given routemap
func validateRoutemap(s *routemap, config *Config) (*routemap, error) {
	// routemap must have at least one variant
	if len(s.Variants) == 0 {
		e := errors.New("routemap must at least one variant")
		log.Logger.Error(e)
		return nil, e
	}

	// if !clusterScoped, variant resource namespace should be nil or equal routemap namespace
	if !config.ClusterScoped {
		for _, v := range s.Variants {
			for _, r := range v.Resources {
				if r.Namespace != nil && *r.Namespace != s.Namespace {
					e := errors.New("expected variant resource namespace to match routemap namespace")
					log.Logger.Error(e)
					return nil, e
				}
			}
		}
	}

	return s, nil
}
```

controllers/config.go
sriumcp: Yes, used by tests in `cmd` package.

sriumcp: No. Fixed.

controllers/podname.go
sriumcp: Fixed.

controllers/podname.go
sriumcp: Fixed.

controllers/podname.go
sriumcp: Fixed.

controllers/routemap.go
sriumcp: This is how the `routemap templates` get weights for variants.

sriumcp: Documented now.

sriumcp:
> I have a general question about the content of testdata/controllers. We have a number of folders with scripts for various kserve use cases. None of this serves the purpose of test data. I am wondering if this is the right place for this. If so, shouldn't the modelmesh versions also be included (there is a PR on the branch for this). I think the right place is probably in the docs project under samples.

Perhaps this needs to be handled in a separate issue? Just so we don't break any existing functionality accidentally?

testdata/controllers/config.yaml
sriumcp: schema.GroupVersionResource is directly embedded in our GVRK struct.

Unfortunately, `schema.GroupVersionResource` has no JSON tags, so they are capitalized.
***
For now, I suggest we retain this approach. The Helm chart can make the assumption that GVR is capitalized, it doesn't need to assume it is small.

testdata/controllers/config.yaml
sriumcp: Fixed.

kalantar (approved):

kalantar (approved):

PR #1442, opened 2023-04-11
replace variant with version

### LINKED ISSUE #1440: Rename variants as versions
We use the term `variants` in routemaps. Rename to `versions`. This seems both clearer and more common use.

kalantar:
Replace the term `variant` with `version`. Fixes https://github.com/iter8-tools/iter8/issues/1440.

### DIFF
--- a/controllers/allcontrollers_test.go
+++ b/controllers/allcontrollers_test.go
@@ -43,7 +43,7 @@ func TestStart(t *testing.T) {
 		Immutable: base.BoolPointer(true),
 		Data: map[string]string{
 			"strSpec": `
-variants:
+versions:
 - resources:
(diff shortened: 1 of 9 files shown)


sriumcp (approved):
lgtm

kalantar:
Why is testing so inconsistent on code coverage?  Coverage drops because of a name change? Even the same number of characters!

sriumcp:
> Why is testing so inconsistent on code coverage? Coverage drops because of a name change? Even the same number of characters!

Didn't realize the coverage was by character count... good to know... 

The previous version would have been borderline (just above `76.1`). This version also could be borderline (just below `76.1`).

PR #1445, opened 2023-04-12
update actions with newer versions

### LINKED ISSUE #1444: Update workflows to eliminate use of node12
Our "Publish binaries and brew formula" workflow has the following warning:
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: peter-evans/repository-dispatch@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

We should update this workflow (and any others with Node.js 12)

kalantar:
Update actions in workflows with newer versions. 

This change `peter-evans/repository-dispatch@v1` --> `peter-evans/repository-dispatch@v2` fixes https://github.com/iter8-tools/iter8/issues/1444

Other updates were made just to keep current

### DIFF
--- a/.github/workflows/assets.yaml
+++ b/.github/workflows/assets.yaml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Install Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@v4
       with:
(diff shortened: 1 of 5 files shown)


sriumcp (approved):
lgtm

PR #1459, opened 2023-05-10
Merge `hub` repo into `iter8` repo

### LINKED ISSUE #1457: Merge `hub` repo into `iter8` repo
Excerpt from #1448:

Merge `hub` repo into `iter8` repo. This way, we will still have a `helm` repo for installing Iter8 helm charts (and not a Git repo). However, we will not have two repos, only one. So, `iter8` will become a mono repo, containing both `Go` code, and `Helm` charts.

Alan-Cha:
Fix #1457

### DIFF
--- a/.github/workflows/release-drafter.yaml
+++ b/.github/workflows/release-drafter.yaml
@@ -9,9 +9,9 @@ jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
-      # Drafts your next Release notes as Pull Requests are merged into any tracked branch
-      - uses: release-drafter/release-drafter@v5
-        with:
-          config-name: release-config.yaml
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    # Drafts your next Release notes as Pull Requests are merged into any tracked branch
+    - uses: release-drafter/release-drafter@v5
+      with:
+        config-name: release-config.yaml
+      env:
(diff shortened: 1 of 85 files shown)


Alan-Cha:
The `Additional Helm chart linting` workflow should fail because of #1460

Alan-Cha:
`Tests / unit-test` workflow should fail because the failing test `TestKubeLaunch` tries to reach the Helm chart in the `iter8` repo, which will be created upon accepting this PR

```
looks like "https://iter8-tools.github.io/iter8" is not a valid chart repository or cannot be reached
```

.github/workflows/release-drafter.yaml
```diff
@@ -1,17 +0,0 @@
-name: Release Drafter
```
sriumcp, line 1: I don't think this file should be removed. It has to do with the release of the Iter8 CLI as well.

sriumcp (changes requested):

.github/workflows/release-drafter.yaml
Alan-Cha: @sriumcp I did not remove this file, I just renamed it. [See here](https://github.com/iter8-tools/iter8/pull/1459/files#diff-5c6af029b7e05efb1bde5bf14595b487ca302b580d7b07deea059ead3d1d8a1a)

I just thought the name was ugly because none of the other workflows have a hyphen in their names

sriumcp (approved):
lgtm

PR #1461, opened 2023-05-11
Update charts

### LINKED ISSUE #1457: Merge `hub` repo into `iter8` repo

Alan-Cha:
See https://github.com/iter8-tools/iter8/issues/1457#issuecomment-1544188549

Fix #1457

### DIFF
--- a/.github/workflows/testkustomize.yaml
+++ b/.github/workflows/testkustomize.yaml
@@ -2,8 +2,6 @@ name: Test kustomize experiments
 
 on:
   pull_request:
-    branches:
-    - master
 
 # Kind versions used to test Iter8 on different versions of Kubernetes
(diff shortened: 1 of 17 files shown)


kalantar:
The failures make sense - I did remove the namespace. If it is required, it should cause a failure. I do not understand why it did not on the hub repo. In any case, it is not required for the namespace based deployment. The value that should be assigned is uncertain to me. I believe it was `default` but this is not valid if the users kubeconfig sets the namespace to something else (as I have been doing).

PR #1468, opened 2023-05-16
Test asset workflow

Alan-Cha:


charts/iter8/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: iter8
-version: 0.13.8
+version: 0.14.0
```
Alan-Cha, line 3: Should I bump the minor version? Because there wasn't much change to the chart itself, I only bumped the patch version.

However, if I understand correctly, `iter8` is the [default chart](https://github.com/iter8-tools/iter8/blob/8f80f45039c32a5d2261a2c47a52bacd9226d55f/action/launch.go#L16) and `0.14.x` is the [default version](https://github.com/iter8-tools/iter8/blob/8f80f45039c32a5d2261a2c47a52bacd9226d55f/action/launch.go#L54) because I bumped [MajorMinor](https://github.com/iter8-tools/iter8/blob/master/base/util.go#L19), so the chart version should match? 

Currently, when I try to run `iter8 k launch`, it tells me that it cannot find the appropriate version:

```
ERRO[2023-05-16 08:31:36] chart "iter8" version "0.14.x" not found in https://iter8-tools.github.io/iter8 repository 
```

I did not know that these were intertwined?

sriumcp: Yes, `0.14.0` seems correct. Chart and iter8 versions are indeed intertwined.

sriumcp (changes requested):

PR #1470, opened 2023-05-16
Revert Dockerfile to use binaries

Alan-Cha:


.github/workflows/assets.yaml
```diff
@@ -1,12 +1,59 @@
...
+        echo "SHAFORMAC=$(grep darwin ../checksum.txt | awk '{print $1}')" >> $GITHUB_ENV
+    - name: Upload checksum to release
+      uses: svenstaro/upload-release-action@v2
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        asset_name: checksum.txt
+        file: checksum.txt
+        tag: ${{ github.ref }}
+        overwrite: true
+
```
Alan-Cha, line 53: Removed `homebrew` section

.github/workflows/assets.yaml
```diff
@@ -1,12 +1,59 @@
...
   build-and-push:
+    name: Push Iter8 image to Docker Hub
+    needs: assets
```
Alan-Cha, line 56: This now requires `assets` job

.github/workflows/assets.yaml
```diff
@@ -40,10 +87,11 @@ jobs:
           TAG=v${{ env.VERSION }}
 
   kubernetes-http-experiment:
-    name: Kubernetes HTTP load test    
-    runs-on: ubuntu-latest
-    needs: 
+    name: Kubernetes HTTP load test
+    needs:
+    - assets
```
Alan-Cha: Tests jobs need both `assets` and `build-and-push` jobs

Dockerfile
```diff
@@ -1,10 +1,20 @@
...
 
 # Install curl
-RUN apt-get update && apt-get install -y golang-go
+RUN apt-get update && apt-get install -y curl
 
-# Install Iter8
-RUN GOBIN=/usr/local/bin go install github.com/iter8-tools/iter8@v0.14
+# Set Iter8 version from build args
+ARG TAG
+ENV TAG=${TAG:-v0.14.0}
```
Alan-Cha, line 9: Bump version to `v0.14.0`

.github/workflows/assets.yaml
sriumcp: Maybe we can remove `assets` from test jobs, since `build-and-push` needs it anyway?

.github/workflows/assets.yaml
```diff
@@ -1,12 +1,59 @@
...
+    - name: Install Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: 1.19
+    - uses: actions/checkout@v3
+    - name: Build binaries
+      run: |
+        VERSION=${GITHUB_REF#refs/*/}
+        echo "Version: ${VERSION}"
+        make dist
```
sriumcp, line 21: Given that only `iter8-linux-amd64.tar.gz` is needed by the Dockerfile, can we please change the `make dist` target in the Makefile to produce only this binary? This will save a bunch of time.

sriumcp (changes requested):

sriumcp (approved):
lgtm

PR #1471, opened 2023-05-16
Release config fixes

sriumcp:
Release drafter got broken because `charts` also create releases/tags. This PR is an attempt to filter tags better to compute next tag.


kalantar (approved):
lgtm

Issue #1462, opened 2023-05-11
Reuse traffic controller config for ABn controller


Alan-Cha:
Excerpt from https://github.com/iter8-tools/iter8/issues/1448:

Ensure that ABn controller re-uses the same config struct/file as the traffic controller. This is a breaking change since for ABn since the config information it uses today is somewhat different. Installation of ABn will need to change accordingly. Also, it will use the `routemap` weights directly from the traffic controller as part of its lookup call, to support weighted routing.

kalantar:
The ABn code has two parts. 

- A controller that watches change to applications and creates an in memory representation.
- A grpc service that responds to `Lookup`, `WriteMetric` and `GetApplication`

The goal of this issue is to use the traffic manager controller instead of the ABn controller. That is, rewrite the service to use the structures created by the traffic controller instead of the current controller.

kalantar:
duplicate

PR #1475, opened 2023-05-19
Persistent volume

Alan-Cha:
Related to #1462


sriumcp:
1. Please target the traffic controller for this, rather than the ABn chart.
2. You can using dynamic provisioning to allocate persistent volume. You don't deal with PV directly. You only deal with PVC and supply the storage class name. For instance, see: https://docs.openshift.com/container-platform/3.11/install_config/storage_examples/storage_classes_dynamic_provisioning.html
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
 name: iter8-pvc
spec:
 accessModes: # list value you can set
  - ReadWriteOnce # only one pod has access to this 
 resources:
   requests:
     storage: 10Gi # value you can set
 storageClassName: default # most cloud vendors have a default storage class defined # value you can set
```

sriumcp (approved):
lgtm

PR #1481, opened 2023-05-19
Reorganize Kustomize traffic files

### LINKED ISSUE #1476: Reduce duplication in Kustomize traffic files
Currently, the [Kustomize traffic files](https://github.com/iter8-tools/iter8/tree/master/kustomize/traffic) are structured as follows:

* kustomize/traffic
  * base
    * kustomization.yaml
    * ...
  * clusterScoped
    * kustomization.yaml
    * clusterroles.yaml
  * namespaceScoped
    * kustomization.yaml
    * roles.yaml

But this introduces a few problems:

* Because [namespace is added](https://github.com/iter8-tools/iter8/blob/a3c3fcae925387c4aa1567daa060b0aad035bd39/kustomize/traffic/base/kustomization.yaml#L7) via the base `kustomization.yaml` file and Role and RoleBinding (as well as ClusterRole and ClusterRoleBinding) are added in the variants, the manifests added by the variants do not get the `namespace` that is added by the base `kustomization.yaml` file.
* The clusterScoped variant has both namespaces added by the base `kustomization.yaml` file as well as hardcoded namespace in `clusterroles.yaml`. This mismatch can cause issues in the future.

***

Solution:

* kustomize/traffic
  * namespaceScoped
    * kustomization.yaml
    * ...
  * clusterScoped
    * kustomization.yaml
    * clusterroles.yaml

Get rid of the base (turn it into namespaceScoped) and make clusterScoped a variant of it. No longer add manifests in `clusterroles.yaml`, just use overlay.

Alan-Cha:
Fix #1476


kustomize/traffic/clusterScoped/kustomization.yaml
```diff
@@ -1,5 +1,25 @@
...
+      path: /roleRef/kind
+      value: ClusterRole
+  target:
+    kind: RoleBinding
+- patch: |-
+    - op: replace
+      path: /kind
+      value: ClusterRoleBinding
+  target:
+    kind: RoleBinding
```
Alan-Cha, line 27: I had some trouble patching the `kind`. Seems like certain methods of patching do no like the `kind` being changed. This method seems to work, however.

kustomize/traffic/clusterScoped/kustomization.yaml
```diff
@@ -1,5 +1,25 @@
 bases:
-- ../base
+- ../namespaceScoped
```
kalantar, line 3: I think we need to add:

`namespace: default`

kustomize/traffic/namespaceScoped/kustomization.yaml
```diff
@@ -1,5 +1,14 @@
...
-- roles.yaml
\ No newline at end of file
+- configmap.yaml
+- headlessservice.yaml
+- role.yaml
+- rolebinding.yaml
+- serviceaccount.yaml
+- statefulset.yaml
+
+namespace: default
```
kalantar: I think we should remove this line so that it will install in the namespace set in the current kubernetes context.

kustomize/traffic/clusterScoped/kustomization.yaml
kalantar: This will install it everything in the `default` namespace only when `clusterScoped` is used.

kalantar (changes requested):

kalantar (approved):
lgtm

PR #1483, opened 2023-05-23
Add interface

Alan-Cha:


Alan-Cha:
I'm still trying to figure out what is needed

controllers/storageclient/interface.go
```diff
@@ -0,0 +1,12 @@
+// Package storageclient provides the storage client for the controllers package
+package storageclient
+
+// Interface enables interaction with a storage entity
+// Can be mocked in unit tests with fake implementation
+type Interface interface {
+	Start(path string, opts interface{}) error
```
Alan-Cha:
```suggestion
	SetUp(path string, opts interface{}) error
```

controllers/storageclient/interface.go
```diff
@@ -0,0 +1,12 @@
...
+	CreateMetric(name string, value interface{}) error
+	GetMetric(name string) (interface{}, error)
+	DeleteMetric(name string) (interface{}, error)
```
Alan-Cha:
```suggestion
```

controllers/storageclient/simple.go
```diff
@@ -0,0 +1,101 @@
...
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/dgraph-io/badger"
+	"github.com/imdario/mergo"
+)
+
+type Client struct {
+	db interface{}
```
Alan-Cha: BadgerDB type

controllers/storageclient/interface.go
```diff
@@ -0,0 +1,12 @@
...
+	Size() (int, int, error) // current capacity and max capacity TODO: units?
```
Alan-Cha: Another function to get the database object

Alan-Cha: GetClient()

controllers/storageclient/simple.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// Start initializes BadgerDB instance
+func (cl *Client) Start(path string, opts interface{}) error {
```
Alan-Cha: User would not supply the `path`

controllers/allcontrollers.go
```diff
@@ -31,6 +34,9 @@ const (
 	iter8KindLabel         = "iter8.tools/kind"
 	iter8KindRoutemapValue = "routemap"
 	iter8VersionLabel      = "iter8.tools/version"
+
+	// for persistent volume
+	metricsPath = "/metrics"
```
sriumcp: constant for mount point?

controllers/config.go
```diff
@@ -37,6 +37,12 @@ type Config struct {
 	DefaultResync string `json:"defaultResync,omitempty"`
 	// ClusterScoped is true if Iter8 controller is cluster-scoped
 	ClusterScoped bool `json:"clusterScoped,omitempty"`
+	// Persist is true if Iter8 controller should have a persistent volume
+	Persist bool `json:"persist,omitempty"`
+	// Storage is the minimum amount of space that can be requested for the persistent volume
+	Storage string `json:"storage,omitempty"`
+	// StorageClassName is the provisioner for the persistent volume
+	StorageClassName string `json:"storageClassName,omitempty"`
```
sriumcp: StorageClassName goes away.

Instead of reading in storage info, potential approach is to try the `volume` package from k8s, get a `NewMetricsDu` provider, and use that to calculate metrics. This means, may be, `storage` also goes away.

controllers/storageclient/badgerdb/simple.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+	"os"
+
+	"github.com/dgraph-io/badger"
+	"github.com/imdario/mergo"
+)
+
+// Client is a client for the BadgerDB
+type Client struct {
+	db *badger.DB
```
sriumcp, line 14: This can get the mount point, may be as a const.

controllers/storageclient/badgerdb/simple.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, errors.New("cannot open BadgerDB")
+	}
+	client.db = db
+
+	return &client, nil
+}
+
+// Size gets the current size and the maximum size of the BadgerDB
+func (cl Client) Size() (int64, int64, error) {
```
sriumcp: Looks like this is changing...

controllers/allcontrollers.go
```diff
@@ -224,6 +230,13 @@ func Start(stopCh <-chan struct{}, client k8sclient.Interface) error {
 		log.Logger.Warnf("could not get pod with name %s in namespace %s", name, ns)
 	}
 
+	if config.Persist {
```
sriumcp, line 235: var dbClient storageclient.Interface

controllers/allcontrollers.go
```diff
@@ -242,3 +257,19 @@ func Start(stopCh <-chan struct{}, client k8sclient.Interface) error {
 
 	return nil
 }
+
+// // getVolumeUsage gets the available and total capacity of a volume, in that order
```
sriumcp: 1. GetVolumeUsage instead of getVolumeUsage
2. var metricsPath instead of const metricsPath
3. Tests (with temp paths and defer cleanup of temp paths) in both the packages.

sriumcp (changes requested):

controllers/allcontrollers.go
Alan-Cha: `metricsPath` was originally suggested to be changed for `const` to `var` so that it can be changed for test cases but it does not seem be necessary.

I did choose to export it though.

controllers/storageclient/badgerdb/simple_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+	"github.com/dgraph-io/badger/v4"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetClient(t *testing.T) {
+	tempDirPath := t.TempDir()
```
Alan-Cha, line 11: It is not necessary to change the directory.

controllers/allcontrollers_test.go
```diff
@@ -128,3 +128,17 @@ routingTemplates:
 		return false
 	}, time.Second*2, time.Millisecond*100)
 }
+
+func TestGetVolumeUsage(t *testing.T) {
+	// GetVolumeUsage is based off of statfs which analyzes the volume, not the directory
+	// Creating a temporary directory will not change anything
+	path, err := os.Getwd()
```
Alan-Cha, line 135: Creating a temporary directory will not do anything here. Unless, we want to create a volume mount? I am not sure if we even have permission to do that.

sriumcp (approved):
lgtm

PR #1485, opened 2023-05-24
check error before add finalizer event

### LINKED ISSUE #1484: Remove AddFinalizer event
Manual stress testing produces the following error on rare occasions:

```
k8s.io/client-go/tools/record.(*recorderImpl).Event(0xc009c16d20?, {0x2bdf0d8?, 0x0?}, {0x27df624?, 0x0?}, {0x280249e?, 0xc0006671f0?}, {0x28252cc?, 0xc0006671f0?})
	k8s.io/client-go@v0.26.3/tools/record/event.go:363 +0x4f
github.com/iter8-tools/iter8/controllers.broadcastEvent({0x2bdf0d8, 0x0}, {0x27df624, 0x6}, {0x280249e, 0x15}, {0x28252cc, 0x22}, {0x2c16f28, 0xc00094f9b0})
	github.com/iter8-tools/iter8/controllers/events.go:23 +0x1de
github.com/iter8-tools/iter8/controllers.addFinalizer.func1()
	github.com/iter8-tools/iter8/controllers/finalizer.go:78 +0xb12
```

**To Reproduce**
Observed running 150 apps (configmaps) each in loop that does:

```
deploy candidate
sleep 5
promote candidate
sleep 5
```

**Expected behavior**
No crash.

kalantar:
Fixes https://github.com/iter8-tools/iter8/issues/1484.

### DIFF
--- a/controllers/finalizer.go
+++ b/controllers/finalizer.go
@@ -62,20 +62,6 @@ func addFinalizer(name string, namespace string, gvrShort string, client k8sclie
 			}).Namespace(u.GetNamespace()).Update(context.TODO(), u, metav1.UpdateOptions{})
 			if e != nil {
 				log.Logger.WithStackTrace(e.Error()).Error("error while updating resource with finalizer")
-			} else {
-				// broadcast event
-
-				// get resource for event broadcasting
-				r, err := client.Resource(schema.GroupVersionResource{
-					Group:    config.ResourceTypes[gvrShort].Group,
-					Version:  config.ResourceTypes[gvrShort].Version,
-					Resource: config.ResourceTypes[gvrShort].Resource,
-				}).Namespace(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-				if err != nil {
-					log.Logger.Warnf("could not get pod with name %s in namespace %s", name, namespace)
-				}
-
-				broadcastEvent(r, corev1.EventTypeNormal, "Added Iter8 finalizer", "Added Iter8 finalizer for resource", client)
 			}
 			return e
 		}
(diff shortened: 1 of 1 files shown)


controllers/finalizer.go
```diff
@@ -62,20 +62,6 @@ func addFinalizer(name string, namespace string, gvrShort string, client k8sclie
...
-			} else {
-				// broadcast event
-
-				// get resource for event broadcasting
-				r, err := client.Resource(schema.GroupVersionResource{
-					Group:    config.ResourceTypes[gvrShort].Group,
-					Version:  config.ResourceTypes[gvrShort].Version,
-					Resource: config.ResourceTypes[gvrShort].Resource,
-				}).Namespace(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-				if err != nil {
```
sriumcp, line 74: Would nesting the broadcast in an else clause solve the issue?

```go
				if err != nil {
 					log.Logger.Warnf("could not get resource with name %s in namespace %s", name, namespace)
 				} else {
   				  broadcastEvent(r, corev1.EventTypeNormal, "Added Iter8 finalizer", "Added Iter8 finalizer for resource", client)
                                }
```

This way, we still get to keep the finalizer event.

kalantar: This might be sufficient. But do we need the event? The failed case is more relevant than the success case.

sriumcp: The event is useful because it describes changes made by Iter8 to an end-user application resource. They created a resource (without a finalizer), and Iter8 modified it by adding a finalizer.
***
So, for whatever reason, their `kubectl delete ns` is hanging, they know why (from the event).

sriumcp (approved):
lgtm

PR #1486, opened 2023-05-30
Add methods to storageclient

Alan-Cha:


controllers/storageclient/interface.go
```diff
@@ -4,7 +4,9 @@ package storageclient
 // Interface enables interaction with a storage entity
 // Can be mocked in unit tests with fake implementation
 type Interface interface {
-	// CreateMetric(name string, value interface{}) error
-	// GetMetric(name string) (interface{}, error)
-	// DeleteMetric(name string) (interface{}, error)
+	WriteMetric(applicationName string, version string, user string, metricName string, metricValue float64) error
+	GetMetric(applicationName string, version string, user string, metricName string) (float64, error)
```
sriumcp:
```go
// map key is a metric name; map value is a summary metric value
type versionMetricSummary map[string]summaryMetric
// GetSummaryMetrics returns versionMetricSummary for each version of the application
GetSummaryMetrics(applicationName string) (map[int]versionMetricSummary, error)
```

controllers/storageclient/interface.go
```diff
@@ -4,7 +4,9 @@ package storageclient
...
+	DeleteMetric(applicationName string, version string, user string, metricName string) (float64, error)
```
sriumcp: We don't need this API, because you have `FreeSpace` API.

controllers/storageclient/interface.go
```diff
@@ -4,7 +4,9 @@ package storageclient
 // Interface enables interaction with a storage entity
 // Can be mocked in unit tests with fake implementation
 type Interface interface {
-	// CreateMetric(name string, value interface{}) error
-	// GetMetric(name string) (interface{}, error)
-	// DeleteMetric(name string) (interface{}, error)
+	WriteMetric(applicationName string, version string, user string, metricName string, metricValue float64) error
```
sriumcp: // WriteMetric is called by the ABn SDK gRPC API implementation. Here, we mean SDK for application clients.

controllers/storageclient/interface.go
sriumcp: // GetSummaryMetrics is called by the gRPC SDK used as part of experimentation. So, the client here, is an Iter8 experiment.

controllers/storageclient/interface.go
```diff
@@ -4,7 +4,9 @@ package storageclient
...
+	GetMetric(applicationName string, version string, user string, metricName string) (float64, error)
+	DeleteMetric(applicationName string, version string, user string, metricName string) (float64, error)
+
+	FreeSpace(bytes uint64) error
```
sriumcp: Ultimately, this method will be invoked by a user. Perhaps, through a REST API.

sriumcp (changes requested):

Alan-Cha:
Questions to be answered:

* How to store data?
  * Serialize the application name/application version/user name/metric name?
    * Use [prefix](https://dgraph.io/docs/badger/get-started/#prefix-scans) scan to get a particular application/application version?
  * Multiple databases for each application? For each application + application version?
  * Use one byte [metadata](https://dgraph.io/docs/badger/get-started/#setting-time-to-live-ttl-and-user-metadata-on-keys)?
    * Metadata can be the version number, either for one application or across all applications
* How to delete data?
  * Set [TTL](https://dgraph.io/docs/badger/get-started/#setting-time-to-live-ttl-and-user-metadata-on-keys)?
    * How long should the TTL be?
    * Should this be user configurable?
  * If we are using multiple databases, what should be the life cycle for each one?
  
***

TTL should be user configurable (maybe default of 2 weeks) keep it long enough for the experiment but not needlessly longer

controllers/storageclient/interface.go
```diff
@@ -1,10 +1,18 @@
...
+type versionMetricSummary map[string]summarymetrics.SummaryMetric
+
 // Interface enables interaction with a storage entity
 // Can be mocked in unit tests with fake implementation
 type Interface interface {
-	// CreateMetric(name string, value interface{}) error
-	// GetMetric(name string) (interface{}, error)
-	// DeleteMetric(name string) (interface{}, error)
+	// WriteMetric is called by the A/B/n SDK gRPC API implementation (SDK for application clients)
+	WriteMetric(applicationName string, version string, user string, metricName string, metricValue float64) error
```
Alan-Cha:
```suggestion
	WriteMetric(applicationName string, version int, signature string, user string, metricName string, metricValue float64) error
```

Alan-Cha: Version number and Kubernetes last modified timestamp can be treated as the actual version

Alan-Cha: Later refinement: instead of using the timestamp, hash the metadata of the Kubernetes object

controllers/storageclient/interface.go
```diff
@@ -1,10 +1,18 @@
...
+
```
Alan-Cha: Key 2: kt-metric::my-app::0::my-signature::my-metric::my-user::my-transaction-id (get/set the metric value with all the provided information)
Key 3: kt-app-metrics::my-app::my-metric -> true (allows us to get metric list for a given app)
Key 5: kt-users::my-app::0::my-signature::my-user -> true (get all users for a particular app version) (getDistinctUserCt())

controllers/storageclient/interface.go
```diff
@@ -1,10 +1,18 @@
 // Package storageclient provides the storage client for the controllers package
 package storageclient
 
+import "github.com/iter8-tools/iter8/base/summarymetrics"
+
+// versionMetricSummary is a map of metric names to summary metric values
+type versionMetricSummary map[string]summarymetrics.SummaryMetric
```
Alan-Cha: Add the number of distinct users to `SummaryMetric`

controllers/storageclient/interface.go
sriumcp: We also want `transactionID string` as input.

```
Key 2: kt-metric::my-app::0::my-signature::my-metric::my-user::my-transaction-ID (the metric value with all the provided information)
```

controllers/storageclient/interface.go
sriumcp: `SetKey5 call is useful` -- will be invoked by lookup.

sriumcp: Support Set for all the above Keys. And support `GetSummaryMetrics`. The latter might need other calls, based on its implementation design (which we're yet to do).

controllers/storageclient/badgerdb/simple.go
```diff
@@ -48,5 +64,256 @@ func GetClient(opts badger.Options) (*Client, error) {
...
+	if err := validateKeyToken(transaction); err != nil {
+		return "", errors.New("transaction ID cannot have \":\"")
+	}
+
+	return fmt.Sprintf("kt-metric::%s::%d::%s::%s::%s::%s", applicationName, version, signature, metric, user, transaction), nil
+}
+
+// GetMetric gets a metric based on the app name, version, signature, metric type, user, and transaction ID from BadgerDB
+// Example key/value: kt-metric::my-app::0::my-signature::my-metric::my-user::my-transaction-id -> my-metric-value
+func (cl Client) GetMetric(applicationName string, version int, signature, metric, user, transaction string) (float64, error) {
```
sriumcp: Where is this used?

controllers/storageclient/badgerdb/simple.go
```diff
@@ -48,5 +64,256 @@ func GetClient(opts badger.Options) (*Client, error) {
...
 	return &client, nil
 }
+
+func getDefaultAdditionalOptions() AdditionalOptions {
+	return AdditionalOptions{
+		TTL: time.Hour * 24,
+	}
+}
+
+func getValueFromBadgerDB(db *badger.DB, key string) ([]byte, error) {
```
sriumcp: Where is this used?

controllers/storageclient/badgerdb/simple.go
Alan-Cha: It is being used in some test cases, I believe.

controllers/storageclient/badgerdb/simple.go
Alan-Cha: It is being used in GetMetric()

controllers/storageclient/badgerdb/simple.go
Alan-Cha: Should I comment this out? The test, as well as `getValueFromBadgerDB()`?

sriumcp: Yes, if a func is getting used only in test cases, we can comment it out. Can be reintroduced later if it is really needed.

sriumcp (changes requested):

controllers/storageclient/badgerdb/simple.go
Alan-Cha: Done.

Alan-Cha:
@sriumcp Commented it out.

sriumcp (approved):
lgtm

PR #1487, opened 2023-05-30
fix for non-default namespaces

kalantar:
Use `.Release.Namespace` as default instead of `modelmesh-serving`.
Modify pod security context for controller.


charts/traffic-templates/templates/_routemap-bluegreen.tpl
```diff
@@ -55,7 +55,7 @@ data:
               {{- range $i, $v := (rest $versions) }}
               {{ `{{- if gt (index .Weights ` }}{{ print (add1 $i) }}{{ `) 0 }}`}}
               - destination:
-                  host: {{ $.Values.modelmeshServingService }}.{{ $.Values.modelmeshServingNamespace }}.svc.cluster.local
+                  host: {{ $.Values.modelmeshServingService }}.{{ $.Release.Namespace }}.svc.cluster.local
```
sriumcp, line 58: Shouldn't this be `modelmeshServingNamespace`? 

***

We have three namespaces. 
1. The namespace where the modelmesh service/controller resides.
2. The namespace where the user's ML model resides.
3. The namespace where the Iter8 controller resides.

Conceptually, variable 3 above has no role in this chart. But it seems like 1 and 2 have roles, and they need to be called out distinctly?

kustomize/traffic/namespaceScoped/statefulset.yaml
```diff
@@ -38,8 +38,12 @@ spec:
             memory: 64Mi
         securityContext:
           readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
           runAsNonRoot: true
-          runAsUser: 1000
+          runAsUser: 1001040000
```
sriumcp, line 46: Could you please provide a reference for this? Where did we pick this value from?

charts/traffic-templates/templates/_routemap-bluegreen.tpl
kalantar: Agreed. However, modelmesh serving deploy a Service `modelmesh-serving` in every namespace enabled for models. To send inference requests to the models in ns `foo`, one sends inference requests to `modelmesh-serving.foo`.  We do want the ns where the user's ML model resides.

kustomize/traffic/namespaceScoped/statefulset.yaml
kalantar: I used the value used by the model mesh pods. I have no other reference. Initially I got an error message that the value 1000 was not allowed (low numbers invalid on openshift)

sriumcp (approved):
lgtm

PR #1490, opened 2023-06-07
GetSummaryMetrics()

Alan-Cha:


controllers/storageclient/badgerdb/simple.go
```diff
@@ -126,23 +133,199 @@ func (cl Client) SetMetric(applicationName string, version int, signature, metri
...
+				tokens := strings.Split(key[len(prefix):], "::")
+				if len(tokens) != 3 {
+					return fmt.Errorf("incorrect number of tokens in metric key \"%s\"; expected 7 tokens", key)
+				}
+
+				currentMetric := tokens[0]
+				currentUser := tokens[1]
+
+				// initial case
+				if previousMetric == "" {
```
sriumcp, line 232: Is there a guarantee that users are going to be grouped together in this for loop?

What if you have:

```
user1
user2
user2
user1
user3
user2
user1
```

Alan-Cha: According to the docs, yes. It says that keys are lexicographically sorted as you iterate through them.

sriumcp (approved):
lgtm

PR #1494, opened 2023-06-10
Set default release in Dockefile to one with existing binaries

### LINKED ISSUE #1488: Using iter8 action fails because the Dockerfile references the wrong GitHub release URL
**Describe the bug**

Building the container image via the Dockerfile fails with `tar: This does not look like a tar archive`. For example, this happens when using this repo as an action in a GitHub workflow.

This happens because release v0.14.0 does not include any binaries as part of the release (https://github.com/iter8-tools/iter8/releases/tag/v0.14.0), but the Dockerfile https://github.com/iter8-tools/iter8/blob/master/Dockerfile#L12 attempts to download the binary anyways.

**To Reproduce**
Run `docker build .` in the root of this repository.

**Expected behavior**
The container should build successfully.

**Screenshots**

image

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS]
 - Output of the `iter8 version` command

**Additional context**

chgl:
Fixes #1488

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && apt-get install -y curl
 
 # Set Iter8 version from build args
 ARG TAG
-ENV TAG=${TAG:-v0.14.0}
+ENV TAG=${TAG:-v0.14.10}
 
 # Download iter8 compressed binary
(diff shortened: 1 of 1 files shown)


sriumcp (approved):
lgtm

PR #1495, opened 2023-06-12
Add signature

Alan-Cha:
Related to #1478

PR #1498, opened 2023-06-14
Add signature

Alan-Cha:
Refactor #1495

### DIFF
--- a/controllers/routemap.go
+++ b/controllers/routemap.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"bytes"
-	"context"
 	"errors"
 	"fmt"
 	"html/template"
@@ -12,11 +11,11 @@ import (
 
 	"github.com/iter8-tools/iter8/base/log"
 	"github.com/iter8-tools/iter8/controllers/k8sclient"
+	"github.com/mitchellh/hashstructure/v2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	seriyaml "k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
@@ -70,6 +69,9 @@ const (
 	routemapStrSpec      = "strSpec"
 	weightAnnotation     = "iter8.tools/weight"
 	defaultVersionWeight = uint32(1)
+	status               = "status"
+	metadata             = "metadata"
+	resourceVersion      = "resourceVersion"
 )
 
 // Weights provide the relative weights for traffic routing between versions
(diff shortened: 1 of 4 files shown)


controllers/routemap_test.go
```diff
@@ -214,3 +217,136 @@ func TestGetObservedGeneration(t *testing.T) {
...
+	o runtime.Object
+}
+
+func (gl testGenericLister) List(selector labels.Selector) (ret []runtime.Object, err error) {
+	return nil, nil
+}
+
+func (gl testGenericLister) Get(name string) (runtime.Object, error) {
+	return gl.o, nil
+}
```
Alan-Cha: This is for mocking the app informer used in `computeSignature()`. This seems to be the best way to unit test this function for now.

controllers/routemap.go
```diff
@@ -216,6 +221,36 @@ versionLoop:
...
+		obj, err := appInformers[resource.GVRShort].Lister().ByNamespace(*resource.Namespace).Get(resource.Name)
+		if err != nil {
+			return 0, fmt.Errorf("cannot get resource: %e", err)
+		}
+
+		// TODO: NestedFieldCopy() or NestedFieldNoCopy()?
+		// extract spec section from resource, if applicable
+		specSection, _, err := unstructured.NestedFieldNoCopy(obj.(*unstructured.Unstructured).Object, spec)
+		if err != nil {
+			return 0, fmt.Errorf("cannot traverse resource: %e", err)
```
kalantar, line 240: Maybe just a warning?

controllers/routemap.go
```diff
@@ -67,6 +69,10 @@ const (
 	routemapStrSpec      = "strSpec"
 	weightAnnotation     = "iter8.tools/weight"
 	defaultVersionWeight = uint32(1)
+	spec                 = "spec"
+	status               = "status"
+	metadata             = "metadata"
+	resourceVersion      = "resourceVersion"
```
kalantar: we don't need these anymore

controllers/routemap.go
kalantar: nevermind

controllers/routemap.go
```diff
@@ -216,6 +221,36 @@ versionLoop:
...
+func computeSignature(v version) (uint64, error) {
+	// get all resources of a version
+	resources := []interface{}{}
+	for _, resource := range v.Resources {
+		obj, err := appInformers[resource.GVRShort].Lister().ByNamespace(*resource.Namespace).Get(resource.Name)
+		if err != nil {
+			return 0, fmt.Errorf("cannot get resource: %e", err)
+		}
+
+		// TODO: NestedFieldCopy() or NestedFieldNoCopy()?
```
kalantar: NestedFieldNoCopy should be fine; we have a copy of the object no reason to copy again

controllers/routemap.go
```diff
@@ -225,6 +260,17 @@ func (s *routemap) reconcile(config *Config, client k8sclient.Interface) {
 	// normalize version weights
 	s.normalizeWeights(config)
 
+	// calculate the signature for the version
+	for _, version := range s.Versions {
+		signature, err := computeSignature(version)
+		if err != nil {
+			// TODO: throw error?
+			log.Logger.Error("cannot calculate signature for version: ", version)
```
kalantar: just return; maybe print error with stacktrace

kalantar (changes requested):

controllers/routemap.go
```diff
@@ -32,7 +33,8 @@ type routemap struct {
 
 type version struct {
 	Resources []resource `json:"resources,omitempty"`
-	Weight    *uint32    `json:"weight,omitempty"`
+	Weight    *uint32    `json:"label,omitempty"`
+	Signature uint64
```
kalantar: need go tag `json`
if no resources, then no signature so use `*unit64`

kalantar (dismissed):
looks good to me

controllers/routemap.go
```diff
@@ -32,7 +33,8 @@ type routemap struct {
 
 type version struct {
 	Resources []resource `json:"resources,omitempty"`
-	Weight    *uint32    `json:"weight,omitempty"`
+	Weight    *uint32    `json:"label,omitempty"`
```
sriumcp: This seems like a bug? `json:"weight,...`

controllers/routemap.go
```diff
@@ -216,6 +218,36 @@ versionLoop:
...
+		}
+
+		// extract spec section from resource, if applicable
+		specSection, _, err := unstructured.NestedFieldNoCopy(obj.(*unstructured.Unstructured).Object, spec)
+		if err != nil {
+			// error here does not mean no spec, it means cannot traverse the object (to find if there is a spec)
+			return 0, fmt.Errorf("cannot traverse resource: %e", err)
+		}
+
+		resources = append(resources, specSection)
```
sriumcp, line 243: Is the length of `resources` always guaranteed to be equal to length of `v.Resources`? This is desirable.

In particular, if some objects don't have spec, it looks like some nil values would be added to `resources` ... correct?

sriumcp (changes requested):
Minor change + q.

controllers/routemap.go
Alan-Cha: Oh, that is strange. I think this may have come from when I was rebasing.

controllers/routemap.go
Alan-Cha: Yes, should be the same length, and yes, `nil` should be added.

controllers/routemap_test.go
```diff
@@ -214,3 +218,214 @@ func TestGetObservedGeneration(t *testing.T) {
 		assert.Equal(t, tt.ok, o)
 	}
 }
+
+type testInformer struct {
```
sriumcp: We have a standard way to mock the Kubernetes client, populate it with objects, and call its APIs: please see [here](https://github.com/iter8-tools/iter8/blob/master/controllers/k8sclient/fake/simple_test.go).

Could we please use the fake client instead of creating our own Kubernetes mocking mechanisms?

sriumcp: A more end-to-end example of how fake client is used is here: https://github.com/iter8-tools/iter8/blob/master/controllers/allcontrollers_test.go

sriumcp (changes requested):
Please see a change requested re: Kubernetes mock.

sriumcp (approved):
lgtm

PR #1511, opened 2023-06-26
GetMetrics() uses app, user, trans

### LINKED ISSUE #1506: GetMetrics() for a particular app/version/signature
Current `GetMetrics()` (in the context of storage client) returns all metrics in the storage client. This may result in a lot of data and it can be simplified.

Change `GetMetrics()` so it uses the app name, version number, and signature to return a subset of the data.

Alan-Cha:
Fix #1506


controllers/storageclient/badgerdb/simple.go
```diff
@@ -287,29 +287,23 @@ func (cl Client) GetSummaryMetrics(applicationName string, version int, signatur
...
-//				}
+//		"my-metric": {
+//			"my-user": {
+//				"my-transaction-id": 5.0
 //			}
 //		}
 //	}
-func (cl Client) GetMetrics() (map[string]map[string]map[string]map[string]map[string]map[string]float64, error) {
-	metrics := map[string]map[string]map[string]map[string]map[string]map[string]float64{}
+func (cl Client) GetMetrics(applicationName string, version int, signature string) (map[string]map[string]map[string]float64, error) {
```
sriumcp:
```go
type metricsVector struct {
  ByUser []float64 `json:"byUser,omitempty"`
  ByTransaction[] float64 ...
}

func (cl Client) GetMetrics(applicationName string, version int, signature string) (metricsVector, error) {
}
```

sriumcp (dismissed):
1. Please return lists
2. Please invoke `GetMetrics` as part of `GetSummaryMetrics`
3. Metrics (HTTP) service (that is yet to be written by Michael) will stitch up all the data for a given app and return as part of the data source query.

controllers/storageclient/badgerdb/simple.go
```diff
@@ -179,234 +176,153 @@ func (cl Client) getUserCount(applicationName string, version int, signature str
...
+// NOTE: for users that have not produced any metrics (for example, via lookup()), GetMetrics() will add 0s for the extra users in metricsOverUsers
+//
+// Example, given 5 total users:
+//
+//	{
+//		"my-metric": {
+//			"MetricsOverTransactions": [1, 1, 3, 4, 5]
+//			"MetricsOverUsers": [2, 7, 5, 0, 0]
+//		}
+//	}
```
Alan-Cha, line 198: This note might need to be documented elsewhere

controllers/storageclient/interface.go
```diff
@@ -56,3 +89,130 @@ type Interface interface {
...
+func roundDecimal(x float64, decimalPlace float64) float64 {
+	y := math.Pow(10, decimalPlace)
+
+	return math.Floor(x*y) / y
+}
+
+// bucketLabel return a label for a histogram bucket
+func bucketLabel(min, max float64, decimalPlace float64) string {
+	return fmt.Sprintf("%s - %s", strconv.FormatFloat(roundDecimal(min, decimalPlace), 'f', -1, 64), strconv.FormatFloat(roundDecimal(max, decimalPlace), 'f', -1, 64))
+}
```
kalantar: We agreed that this should go in the (HTTP) metrics service. I have a copy of this already. Let's remove it from here now.

controllers/storageclient/badgerdb/simple.go
kalantar: @Alan-Cha You documented this in the interface. Perhaps that is sufficient. Having it here again isn't a big deal.

controllers/storageclient/interface_test.go
```diff
@@ -0,0 +1,135 @@
+package storageclient
```
kalantar: similarly, these tests should be removed from here (I've made a copy)

kalantar (changes requested):
please remove parts that we decided would go to (http) metrics service

kalantar (dismissed):

sriumcp (approved):
lgtm

PR #1543, opened 2023-07-14
update to v0.15

kalantar:
Update charts and tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && apt-get install -y curl
 
 # Set Iter8 version from build args
 ARG TAG
-ENV TAG=${TAG:-v0.15.1}
+ENV TAG=${TAG:-v0.15.0}
 
 # Download iter8 compressed binary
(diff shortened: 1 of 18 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update && apt-get install -y curl
 
 # Set Iter8 version from build args
 ARG TAG
-ENV TAG=${TAG:-v0.14.10}
+ENV TAG=${TAG:-v0.15.1}
```
Alan-Cha: I notice that this is `v0.15.1`...

charts/iter8/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: iter8
-version: 0.14.3
+version: 0.15.0
```
Alan-Cha, line 3: ... and this is `v0.15.0`. Is this what we want?

Dockerfile
kalantar: fixed

charts/iter8/Chart.yaml
kalantar: we want this one

Alan-Cha (approved):

PR #1571, opened 2023-08-08
Add http payload and http/grpc multiple tests to workflow 

Alan-Cha:
Fix https://github.com/iter8-tools/iter8/issues/1566

Related to https://github.com/iter8-tools/iter8/issues/1567

PR #1577, opened 2023-08-13
Test gRPC workflows with double quotes

Alan-Cha:
Related to #1571


Alan-Cha:
Double quotes does not seem to work as expected.

Alan-Cha:
Double quotes seem to work

PR #1615, opened 2023-08-31
Delete extraneous files

Alan-Cha:

PR #1622, opened 2023-09-05
Various changes

### LINKED ISSUE #1492: Remove public facing `iter8` commands
Remove unused iter8 commands:

- [ ] `iter8 k launch` (replaced by `helm upgrade [--install]`)
- [ ] `iter8 k log` (replaced by `kubectl logs`)
- [ ] `iter8 k delete` (replaced by `helm delete`)

### LINKED ISSUE #1598: Bump go version for workflows
Workflows are currently referencing go version `1.19` however the latest version is `1.21`.

### LINKED ISSUE #1611: Remove duplicate payload templates for `http` and `slack`
The same payload templates for `http` and `slack` exists in both `charts/iter8` and `templates/notify`. We should remove one.

Generally leaning towards removing `charts/iter8`. In addition, we also need to verify the default `payloadTemplateURL` in both the `slack` and `github` tasks as well as in the docs.

### LINKED ISSUE #1617: Align iter8 to match terminology used in docs
With v0.16, a number of concepts have been removed, which have been eliminated in the docs but not in iter8

For example:

* The metric server expects an `experiment` query parameter for both the HTTP and gRPC dashboards, but the term `experiment` has been removed from the documentation. Some alternatives would be `test` or `performanceTest`.

  Query string: `namespace=default&experiment=default`

* The `slack` and `github` tasks (which under the covers is the `notify` task) both use a payload template that has access to a `.Report` field. The documentation makes no mention of this. The documentation makes mention of an experiment summary, however.

Alan-Cha:
Fix #1598
Fix #1615 
Fix #1611 
Fix #1617 
Fix #1492


cmd/k.go
```diff
@@ -15,7 +15,8 @@ var kcmd = &cobra.Command{
 	Long:  "Work with Kubernetes experiments",
 }
 
-// addExperimentGroupFlag adds the experiment group flag
+// TODO: is group still needed?
+// addExperimentGroupFlag adds the test group flag
 func addExperimentGroupFlag(cmd *cobra.Command, groupP *string) {
 	cmd.Flags().StringVarP(groupP, "group", "g", driver.DefaultExperimentGroup, "name of the experiment group")
 }
```
Alan-Cha: Group is now the name of the experiment right? Does this need to be renamed?

cmd/krun.go
```diff
@@ -23,7 +23,7 @@ func newKRunCmd(kd *driver.KubeDriver, out io.Writer) *cobra.Command {
 	actor.EnvSettings = settings
 	cmd := &cobra.Command{
 		Use:          "run",
-		Short:        "Run a Kubernetes experiment",
+		Short:        "Run a Kubernetes test",
```
Alan-Cha, line 26: Revert these changes

cmd/k.go
kalantar: Agree that this flag is no longer needed.

cmd/krun.go
```diff
@@ -10,11 +10,11 @@ import (
 
 // krunDesc is the description of the k run command
 const krunDesc = `
-Run a Kubernetes experiment. This command reads an experiment specified in a secret and writes the result back to the secret.
+Run a Kubernetes test. This command reads an test specified in a secret and writes the result back to the secret.
```
kalantar: - "Run a performance test on Kubernetes"
- "reads a test" (not an)

cmd/krun.go
kalantar: "Run a performance test"

charts/iter8/README.md
```diff
@@ -1,3 +1,3 @@
-# Iter8 experiment chart
+# Iter8 test chart
```
kalantar: perhaps this is a test "runner" chart?

cmd/krun.go
```diff
@@ -10,11 +10,11 @@ import (
...
 
-	$ iter8 k run --namespace {{ .Experiment.Namespace }} --group {{ .Experiment.group }}
+	$ iter8 k run --namespace {{ .Test.Namespace }} --group {{ .Test.group }}
```
kalantar: remove `--group`

cmd/root.go
```diff
@@ -47,16 +45,6 @@ func Execute() {
 	cobra.CheckErr(rootCmd.Execute())
 }
```
kalantar, line 47: Isn't this still needed to support reading of value.yaml (and overriding it). for iter8 and controller?

cmd/krun.go
kalantar: also change _k-job.tpl to remove this option

cmd/k.go
kalantar: Actually, how is the `--group` option used by `iter8 k run`? Do we still need it for that or can we get it from a name instead?  If not, it might still be needed.

cmd/root.go
Alan-Cha: The linter said that this function was not used so I removed it but I think you are right

Alan-Cha: This function was only used in `iter8 k launch`, not in `iter8 k run`.

base/collect_http.go
```diff
@@ -81,30 +82,6 @@ var (
...
-			return true
-		}
-		// if both limits are present (check both)
-		if lims.Upper != nil && lims.Lower != nil && code <= *lims.Upper && code >= *lims.Lower {
-			return true
-		}
-	}
-	return false
-}
-
```
Alan-Cha, line 107: This is not used

kalantar (dismissed):
looks good

kalantar (commented):
lgtm

PR #1623, opened 2023-09-06
remove iter8 cli from workflows

kalantar:


Alan-Cha (dismissed):

.github/workflows/testcharts.yaml
```diff
@@ -160,10 +153,11 @@ jobs:
         helm install iter8 charts/controller --set logLevel=trace
         kubectl rollout status --watch --timeout=60s statefulset/iter8
 
-    - name: iter8 k launch
+    - name: Start performance experiment
```
Alan-Cha: Performance experiment or performance test? I think it should be the latter, right?

Alan-Cha (approved):

PR #1628, opened 2023-09-07
Modify metric endpoints

### LINKED ISSUE #1620: Proposal: modify metrics endpoints to make more similar
To define data sources we have the following:

- **HTTP**:
  - URL: `http://iter8.default:8080/httpDashboard`
  - query string: `namespace=ns&experiment=test_name`
- **gRPC**:
  - URL: `http://iter8.default:8080/grpcDashboard`
  - query string: `namespace=ns&experiment=test_name`
- **A/B/n**:
  - URL: `http://iter8.default:8080/metrics`
  - query string: `application=ns%2Fapp`

I wondered if we wanted to regularize this:
- For A/B/n use `abnDashboard` instead of `metrics` as the endpoint
- For A/B/n change structure of query string to be `namespace=ns&application=app`
- change `experiement` to `test` or `performance_test`
- Not sure how to make the second parameter the same; but there is no real need to

Alan-Cha:
Fix #1620


Alan-Cha:
See #1629

PR #1629, opened 2023-09-07
Modify metric endpoints

### LINKED ISSUE #1620: Proposal: modify metrics endpoints to make more similar

Alan-Cha:
Fix #1620 

Replace #1628

### DIFF
--- a/base/metrics.go
+++ b/base/metrics.go
@@ -14,11 +14,11 @@ const (
 	// MetricsServerURL is the URL of the metrics server
 	MetricsServerURL = "METRICS_SERVER_URL"
 
-	// MetricsPath is the path to the GET /metrics endpoint
-	MetricsPath = "/metrics"
-
 	// ExperimentResultPath is the path to the PUT /experimentResult endpoint
 	ExperimentResultPath = "/experimentResult"
+
+	// AbnDashboard is the path to the GET /abnDashboard endpoint
+	AbnDashboard = "/abnDashboard"
 	// HTTPDashboardPath is the path to the GET /httpDashboard endpoint
 	HTTPDashboardPath = "/httpDashboard"
 	// GRPCDashboardPath is the path to the GET /grpcDashboard endpoint
--- a/base/metrics.go
+++ b/base/metrics.go
@@ -14,8 +14,8 @@ const (
 	// MetricsServerURL is the URL of the metrics server
 	MetricsServerURL = "METRICS_SERVER_URL"
 
-	// ExperimentResultPath is the path to the PUT /testResult endpoint
-	ExperimentResultPath = "/testResult"
+	// TestResultPath is the path to the PUT /testResult endpoint
+	TestResultPath = "/testResult"
 
 	// AbnDashboard is the path to the GET /abnDashboard endpoint
 	AbnDashboard = "/abnDashboard"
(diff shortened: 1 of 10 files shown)


metrics/server_test.go
```diff
@@ -655,8 +655,8 @@ func TestGetMetrics(t *testing.T) {
 	w := httptest.NewRecorder()
 	rm := allRoutemaps.GetAllRoutemaps().GetRoutemapFromNamespaceName("default", "test")
 	assert.NotNil(t, rm)
-	req := httptest.NewRequest(http.MethodGet, "/metrics?application=default%2Ftest", nil)
-	getMetrics(w, req)
+	req := httptest.NewRequest(http.MethodGet, util.AbnDashboard+"?application=default%2Ftest", nil)
```
Alan-Cha: Hmmm actually this query string was not modified and the test did not fail. Will look into this.

base/metrics.go
```diff
@@ -14,11 +14,11 @@ const (
 	// MetricsServerURL is the URL of the metrics server
 	MetricsServerURL = "METRICS_SERVER_URL"
 
-	// MetricsPath is the path to the GET /metrics endpoint
-	MetricsPath = "/metrics"
-
 	// ExperimentResultPath is the path to the PUT /experimentResult endpoint
 	ExperimentResultPath = "/experimentResult"
```
kalantar: should this also be changed to `testResult` ?

metrics/server.go
```diff
@@ -193,17 +192,27 @@ func getMetrics(w http.ResponseWriter, r *http.Request) {
...
 	}
-	log.Logger.Tracef("getMetrics called for application %s", application)
+
+	namespaceApplication := fmt.Sprintf("%s/%s", namespace, application)
+
+	log.Logger.Tracef("getAbnDashboard called for application %s", namespaceApplication)
 
 	// identify the routemap for the application
-	namespace, name := util.SplitApplication(application)
+	namespace, name := util.SplitApplication(namespaceApplication)
```
kalantar: this shouldn't be necessary any more. namespace name are namespace, application from a few lines above.

kalantar (dismissed):

metrics/server.go
Alan-Cha: Oop sorry

metrics/server.go
```diff
@@ -724,10 +732,10 @@ func getGRPCDashboard(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write(dashboardBytes)
 }
 
-// putExperimentResult handles PUT /experimentResult with query parameter application=namespace/name
+// putExperimentResult handles PUT /testResult with query parameter test=name and namespace=namespace
```
Alan-Cha, line 735: Note gap in function name `putExperimentResult` and endpoint `/testResult with query parameter test=name and namespace=namespace`

metrics/server.go
```diff
@@ -742,13 +750,13 @@ func putExperimentResult(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	experiment := r.URL.Query().Get("experiment")
+	experiment := r.URL.Query().Get("test")
```
Alan-Cha, line 753: Note gap here

kalantar (commented):
lgtm

kalantar (approved):

PR #1636, opened 2023-09-11
modify templates for readiness

kalantar:
Redefine iter8 readiness templates using a map `resourceTypes` similar to that in the controller chart.


kalantar:
replaced by PR  https://github.com/iter8-tools/iter8/pull/1639 and https://github.com/iter8-tools/iter8/pull/1640

PR #1639, opened 2023-09-15
unify specs readiness/controller

kalantar:


Alan-Cha (approved):
It would be nice to see a test with multiple conditions

PR #1640, opened 2023-09-18
Updates to charts for readiness changes

kalantar:
- Update controller chart to simplify `resourceTypes`
- Update iter8 chart to introduce `resourceTypes` matching those in the controller chart
- Update kustomize folder
- Corresponding updates to workflow

Expect chart lint failures since not read only.

### DIFF
--- a/.github/workflows/verifyuserexperience.yaml
+++ b/.github/workflows/verifyuserexperience.yaml
@@ -30,7 +30,7 @@ jobs:
 
     - name: Install controller
       run: |
-        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --version 0.18 --version 0.18 --set logLevel=trace
+        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --version 0.18 --set logLevel=trace
         kubectl rollout status --watch --timeout=60s statefulset/iter8
 
     - name: Start performance test
--- a/.github/workflows/verifyuserexperience.yaml
+++ b/.github/workflows/verifyuserexperience.yaml
@@ -30,13 +30,13 @@ jobs:
 
     - name: Install controller
       run: |
-        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --set logLevel=trace
+        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --version 0.18 --version 0.18 --set logLevel=trace
         kubectl rollout status --watch --timeout=60s statefulset/iter8
 
     - name: Start performance test
       run: |
         helm upgrade --install \
-        --repo https://iter8-tools.github.io/iter8 --version 0.17 httpbin-test iter8 \
+        --repo https://iter8-tools.github.io/iter8 --version 0.18 httpbin-test iter8 \
         --set "tasks={http}" \
         --set http.url="http://httpbin.default/get" \
         --set logLevel=trace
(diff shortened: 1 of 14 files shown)


.github/workflows/verifyuserexperience.yaml
```diff
@@ -30,13 +30,13 @@ jobs:
 
     - name: Install controller
       run: |
-        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --set logLevel=trace
+        helm install --repo https://iter8-tools.github.io/iter8 iter8 controller --version 0.18 --version 0.18 --set logLevel=trace
```
Alan-Cha: version 18 is set twice here?

charts/iter8/templates/_k-role.tpl
```diff
@@ -12,46 +12,28 @@ rules:
...
+{{- range $type, $name := $typesToCheck }}
+{{- $definition := get $.Values.resourceTypes $type }}
+{{- if not $definition }}
+{{- cat "no type definition for: " $type | fail }}
+{{- else }}
+- apiGroups: [ {{ get $definition "Group" | quote }} ]
+  resourceNames: [ {{ $name | quote }} ]
+  resources: [ {{ get $definition "Resource" | quote }} ]
+  verbs: [ "get" ]
+{{- end }} {{- /* if not $definition */}}
```
Alan-Cha, line 36: Are these commented lines necessary?

charts/iter8/templates/_task-ready.tpl
```diff
@@ -1,68 +1,30 @@
...
-    resource: chaosengines
-{{- include "task.ready.tn" . }}
-{{- end }}
-{{- end }}
-{{- end }}
\ No newline at end of file
+    name: {{ $name | quote }}
+    group: {{ get $definition "Group" | quote }}
+    version: {{ get $definition "Version" }}
+    resource: {{ get $definition "Resource" }}
```
Alan-Cha: Some of these fields are quoted while some are not. I see in the previous version, only the name was quoted. Just wondering if we should make these consistent.

charts/iter8/templates/_task-ready.tpl
```diff
@@ -1,68 +1,30 @@
...
+{{- if $namespace }}
+    namespace: {{ $namespace }}
+{{- end }} {{- /* if $namespace */}}
+{{- if $.Values.ready.timeout }}
+    timeout: {{ $.Values.ready.timeout }}
+{{- end }} {{- /* if $.Values.ready.timeout */}}
+{{- end }} {{- /* if not $definition */}}
+{{- end }} {{- /* range $type, $name */}}
+{{- end }} {{- /* {{- if .Values.ready */}}
+{{- end }} {{- /* define "task.ready" */}}
```
Alan-Cha: Commented lines?

kalantar: I'm just commenting what the `{{- end }}` corresponds to. I have trouble following because of the indentation so I've started adding explicit comments when I modify the files.

charts/iter8/templates/_k-role.tpl
kalantar: no. I find them helpful to follow control flow.

charts/iter8/templates/_task-ready.tpl
kalantar: Yeah. Not sure why it was this way. I'll add quote to them all though I think it isn't necessary.

.github/workflows/verifyuserexperience.yaml
kalantar: fixed

Alan-Cha (approved):

PR #1641, opened 2023-09-19
update version

kalantar:


base/util.go
```diff
@@ -19,11 +19,11 @@ import (
 
 // MajorMinor is the minor version of Iter8
 // set this manually whenever the major or minor version changes
-var MajorMinor = "v0.17"
+var MajorMinor = "v0.18"
 
 // Version is the semantic version of Iter8 (with the `v` prefix)
 // Version is intended to be set using LDFLAGS at build time
-var Version = "v0.17.0"
+var Version = "v0.18.1"
```
Alan-Cha: Does this need to be `.1`? We've always left the patch version as `.0`.

charts/routing-actions/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: routing-actions
-version: 0.1.5
+version: 0.18.0
```
Alan-Cha, line 3: This version iv `0.18.0` but I see the version in the controller is `0.18.1`. Is that intentional?

kalantar: There is no reason to keep them in sync at the patch level. This is the first update to the routing-action chart. The controller chart was already at 18.0

base/util.go
kalantar: done

Alan-Cha (approved):

PR #1653, opened 2023-11-08
add redis as alternative metrics store

kalantar:


storage/util.go
```diff
@@ -19,3 +25,71 @@ func GetVolumeUsage(path string) (uint64, uint64, error) {
...
+	return key
+}
+
+// GetExperimentResultKey returns a performance experiment key from the inputs
+func GetExperimentResultKey(namespace, experiment string) string {
+	// getExperimentResultKey() is just getUserPrefix() with the user appended at the end
+	return fmt.Sprintf("kt-result::%s::%s", namespace, experiment)
+}
+
+// GetExperimentResult returns an experiment result retrieved from a key value store
```
Alan-Cha, line 81: From a key value store? Seems like it could be from any store or any function for that matter.

charts/controller/values.yaml
```diff
@@ -57,13 +57,30 @@ resources:
...
+    storage: 50Mi
+    storageClassName: standard
+    # location to mount storage
+    dir: /metrics
+  # properties specific to Redis
+  redis:
+    address: redis:6379
+    # password: (default - none)
+    # username: (default - none)
+    # db: (default 0)
```
Alan-Cha: I wonder if this should be commented. You already provide reasonable defaults for BadgerDB in the Helm charts. It might be a bit ambiguous to the user if there are elements in the `values.yaml` that are not being used in the current Helm deployment. It makes sense to me if the `values.yaml` only contains elements that are directly contributing to he current deployment. I think it's good to make the user aware that they can override the defaults like this but perhaps it should just be a comment here?

storage/redis/simple.go
```diff
@@ -0,0 +1,212 @@
...
+
+	prefix := storage.GetMetricKeyPrefix(applicationName, version, signature)
+	ctx := context.Background()
+	cursor := uint64(0)
+	it := cl.rdb.Scan(ctx, cursor, prefix+"*", int64(0)).Iterator()
+	for it.Next(ctx) {
+		key := it.Val()
+		tokens := strings.Split(key, "::")
+		if len(tokens) != 7 {
+			return nil, fmt.Errorf("incorrect number of toekns in metrics key")
```
Alan-Cha:
```suggestion
			return nil, fmt.Errorf("incorrect number of tokens in metrics key")
```

storage/redis/simple.go
```diff
@@ -0,0 +1,212 @@
...
+type Client struct {
+	rdb *redis.Client
+}
+
+// GetClient returns a Redis client
+func GetClient(config ClientConfig) (*Client, error) {
+	options := &redis.Options{}
+	options.Addr = *config.Address
+	options.Password = "" // default
+	options.DB = 0        //default
```
Alan-Cha:
```suggestion
	options.DB = 0        // default
```

Alan-Cha (dismissed):
Generally looks good!

storage/redis/simple.go
kalantar: done

storage/util.go
kalantar: Possibly.. But we've used it only for key value store so I left the comment specific.

Alan-Cha (dismissed):

Alan-Cha (approved):

PR #1655, opened 2023-11-29
allow num replicas

kalantar:


charts/controller/templates/statefulset.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   {{ template "iter8-controller.labels" . }}
 spec:
   serviceName: {{ .Release.Name }}
+  replicas: {{ default 1 .Values.replicas | int }}
```
Alan-Cha, line 8: Is it strange to provide a default in `values.yaml` as well as a default in the charts? I suppose redundancy does not hurt.

Alan-Cha (approved):

